PR #70, opened 2018-12-08
add @chandanrjit as a team member 

chandanrjit:
Initially forgot to write. My primary focus will be on security on deprecated packages . Most of the organisation who are using nodejs have concerns over security . Would  contribute to make the nodejs ecosystem healthy :)


chandanrjit:
review please !

PR #124, opened 2019-01-07
doc: add doc directory

mhdawson:
We need a place to start adding
documenting that we are working on within
the repository.


mhdawson:
@mcollina could you add your approval to this?

PR #140, opened 2019-01-28
doc: add minutes for meeting 28 January

mhdawson:


dominykas:
Missed the call due to an epic 🤦‍♂️, but I can see a related discussion - coincidentally last week I implemented this: https://www.npmjs.com/package/allow-scripts

tl;dr:

- add [`allowedScripts: { pkg: versionRange }`](https://github.com/yarnpkg/rfcs/pull/76#issuecomment-319614910) in your `package.json`
- `npm install --ignore-scripts` or `yarn install --ignore-scripts`
- `npx allowed-scripts`

fwiw, I hope I can deprecate that package soon, as `--ignore-scripts` + `allowedScripts` should be the default in the package installers.

Not sure how much in scope this is for `package-maintenance`, probably more of a Security WG (will try to find a relevant issue there...)

ljharb:
@dominykas you'd have to pursue that change in both npm and yarn.

dominykas:
Yarn already has an RFC for that, npm has the thread that was mentioned in the call.

I agree security wg doesn't have levers to pull this off, but advocacy and awareness is still useful? And --ignore-scripts as a best practice (at least in CI) is also recommendable?

PR #146, opened 2019-01-31
Create ci-cd-guidelines.md

bnb:
An initial proposal for `ci-cd-automation-guidelines.md`, which outlines bare minimum set of requirements around CI/CD templates.

### DIFF
--- a/docs/drafts/ci-cd-guidelines.md
+++ b/docs/drafts/ci-cd-guidelines.md
@@ -1,5 +1,7 @@
 ## What?
-Guidelines for CI/CD automation templates.
+Guidelines for CI/CD automation templates to be shipped by the package-maintenance team.
+
+> **Note:** This set of guidelines _is not_ about telling projects what they can and cannot do. Instead, it's a framework for those who are interested in building out bare-minium CI/CD templates for projects in the JavaScript ecosystem to use.
 
 ## Why?
 Provide a baseline, common set of CI/CD infrastructure that can be relied upon regardless of what CI/CD providers are used by a project.
@@ -9,30 +11,37 @@ These guidelines are intended to be just that – guidelines. We will also inclu
 
 It's worth noting that projects **do not** need to use one of the provided templates, but **do** need to ensure that their CI/CD setup enables them to follow the required guidelines at the very minimum. Projects are by no means restricted to the required guidelines – more complex and dynamic configurations are more than welcome.
 
-## CI/CD Templates Should Include:
+## Definitions
+Definitions for the various terms used in the guidelines:
 
+### Required
+To say a CI/CD template follows the "tests" guidelines documented here, it would need to meet all of the **required** marks. The same approach applies to for all other guidelines.
+
+### Optional
+**Optional** guidelines are, as the name indicates, optional for templates to include. They're _nice to have_ but by no means required.
+
+## CI/CD Guidelines:
 - Tests
-  - Criteria (**Required**):
+  - **Required**:
     - `npm test`
       - This can be any testing framework exposed via `npm test`.
       - Many CI/CD systems run this implicitly for JavaScript projects. You do not need to explicitly define `npm test` in these cases.
-  - Criteria (**Optional**):
+    - `npm ls`
+      - If `npm ls` exits with a `0` exit code, your build should pass
+      - If `npm ls` exits with a non-`0` exit code, your dependnecy graph is invalid and should fail the run
+  - **Optional**:
     - code coverage
       - failure if code coverage is below a well-documented threshold
-- Caching
-  - Criteria (**Required**):
-    - At a minimum, cache `node_modules` (or equivalent) for your project
 - Operating System Versions
-  - Criteria (**Required**):
+  - **Required**:
     - Windows
     - macOS
     - Linux
-      - At least one current and one LTS release
-      - Ubuntu is suggested but not required
+      - All active LTS releases of Ubuntu or your CI provider's default Linux OS
 - Node.js Versions
-  - Criteria (**Required**):
+  - **Required**:
     - nodejs@current
     - nodejs@lts
       - This includes all active and maintenance versions
-  - Criteria (**Optional**):
+  - **Optional**:
     - nodejs@nightly
--- a/docs/drafts/ci-cd-guidelines.md
+++ b/docs/drafts/ci-cd-guidelines.md
@@ -1,0 +1,38 @@
+## What?
+Guidelines for CI/CD automation templates.
+
+## Why?
+Provide a baseline, common set of CI/CD infrastructure that can be relied upon regardless of what CI/CD providers are used by a project.
+
+## How?
+These guidelines are intended to be just that – guidelines. We will also include templates for common CI/CD providers and accept contributions of templates that implement these guidelines from the community.
+
+It's worth noting that projects **do not** need to use one of the provided templates, but **do** need to ensure that their CI/CD setup enables them to follow the required guidelines at the very minimum. Projects are by no means restricted to the required guidelines – more complex and dynamic configurations are more than welcome.
+
+## CI/CD Templates Should Include:
+
+- Tests
+  - Criteria (**Required**):
+    - `npm test`
+      - This can be any testing framework exposed via `npm test`.
+      - Many CI/CD systems run this implicitly for JavaScript projects. You do not need to explicitly define `npm test` in these cases.
+  - Criteria (**Optional**):
+    - code coverage
+      - failure if code coverage is below a well-documented threshold
+- Caching
+  - Criteria (**Required**):
+    - At a minimum, cache `node_modules` (or equivalent) for your project
+- Operating System Versions
+  - Criteria (**Required**):
+    - Windows
+    - macOS
+    - Linux
+      - At least one current and one LTS release
+      - Ubuntu is suggested but not required
+- Node.js Versions
+  - Criteria (**Required**):
+    - nodejs@current
+    - nodejs@lts
+      - This includes all active and maintenance versions
+  - Criteria (**Optional**):
+    - nodejs@nightly
--- a/docs/drafts/ci-cd-guidelines.md
+++ b/docs/drafts/ci-cd-guidelines.md
@@ -1,26 +1,33 @@
-## What?
+## What
+
 Guidelines for CI/CD automation templates to be shipped by the package-maintenance team.
 
 > **Note:** This set of guidelines _is not_ about telling projects what they can and cannot do. Instead, it's a framework for those who are interested in building out bare-minium CI/CD templates for projects in the JavaScript ecosystem to use.
 
-## Why?
+## Why
+
 Provide a baseline, common set of CI/CD infrastructure that can be relied upon regardless of what CI/CD providers are used by a project.
 
-## How?
+## How
+
 These guidelines are intended to be just that – guidelines. We will also include templates for common CI/CD providers and accept contributions of templates that implement these guidelines from the community.
 
 It's worth noting that projects **do not** need to use one of the provided templates, but **do** need to ensure that their CI/CD setup enables them to follow the required guidelines at the very minimum. Projects are by no means restricted to the required guidelines – more complex and dynamic configurations are more than welcome.
 
 ## Definitions
+
 Definitions for the various terms used in the guidelines:
 
 ### Required
+
 To say a CI/CD template follows the "tests" guidelines documented here, it would need to meet all of the **required** marks. The same approach applies to for all other guidelines.
 
 ### Optional
+
 **Optional** guidelines are, as the name indicates, optional for templates to include. They're _nice to have_ but by no means required.
 
-## CI/CD Guidelines:
+## CI/CD Guidelines
+
 - Tests
   - **Required**:
     - `npm test`
--- a/docs/drafts/ci-cd-guidelines.md
+++ b/docs/drafts/ci-cd-guidelines.md
@@ -35,7 +35,7 @@ To say a CI/CD template follows the "tests" guidelines documented here, it would
       - Many CI/CD systems run this implicitly for JavaScript projects. You do not need to explicitly define `npm test` in these cases.
     - `npm ls`
       - If `npm ls` exits with a `0` exit code, your build should pass
-      - If `npm ls` exits with a non-`0` exit code, your dependnecy graph is invalid and should fail the run
+      - If `npm ls` exits with a non-`0` exit code, your dependency graph is invalid and should fail the run
   - **Optional**:
     - code coverage
       - failure if code coverage is below a well-documented threshold


docs/drafts/ci-cd-guidelines.md
```diff
@@ -0,0 +1,38 @@
...
+  - Criteria (**Required**):
+    - `npm test`
+      - This can be any testing framework exposed via `npm test`.
+      - Many CI/CD systems run this implicitly for JavaScript projects. You do not need to explicitly define `npm test` in these cases.
+  - Criteria (**Optional**):
+    - code coverage
+      - failure if code coverage is below a well-documented threshold
+- Caching
+  - Criteria (**Required**):
+    - At a minimum, cache `node_modules` (or equivalent) for your project
```
ljharb: this is an optimization but doesn't have any impact on the quality of a package. why is this here?

docs/drafts/ci-cd-guidelines.md
```diff
@@ -0,0 +1,38 @@
...
+- Operating System Versions
+  - Criteria (**Required**):
+    - Windows
+    - macOS
+    - Linux
+      - At least one current and one LTS release
```
ljharb: for me, all active LTS releases is a minimum - all releases going back as far as possible is desirable

docs/drafts/ci-cd-guidelines.md
```diff
@@ -0,0 +1,38 @@
...
+      - Ubuntu is suggested but not required
```
ljharb: why does the choice of linux flavor matter?

docs/drafts/ci-cd-guidelines.md
```diff
@@ -0,0 +1,38 @@
...
+- Node.js Versions
+  - Criteria (**Required**):
+    - nodejs@current
+    - nodejs@lts
+      - This includes all active and maintenance versions
+  - Criteria (**Optional**):
+    - nodejs@nightly
```
ljharb: i don't think it's common or even a good idea for most packages to test on nightly; nightlies are for developers on node core, not for ecosystem users.

docs/drafts/ci-cd-guidelines.md
```diff
@@ -0,0 +1,38 @@
...
+## CI/CD Templates Should Include:
+
+- Tests
+  - Criteria (**Required**):
+    - `npm test`
+      - This can be any testing framework exposed via `npm test`.
+      - Many CI/CD systems run this implicitly for JavaScript projects. You do not need to explicitly define `npm test` in these cases.
+  - Criteria (**Optional**):
+    - code coverage
+      - failure if code coverage is below a well-documented threshold
```
wesleytodd: Coverage reports **are** good, but I personally am not a fan of failing.  Code coverage is a notoriously poor indicator of good testing, and suggesting failure on as coverage threshold is a reinforcement of the problem.

I don't know of tooling that does this, but it would be great if you could monitor if it drops significantly (by percentage) from one commit to another. If that existed I would be more willing to recommend warning or failing.

mhdawson: @bcoe just added an option to C8 which allows a fail below a certain level.  We plan to use this in the community regression runs as a "sanity" check, setting it well below the current level. It should also be perfectly acceptable to lower that threshold in a PR as we know coverage goes up/down as new features etc go in.  I see the threshold as a good canary and can be used to achieve what @wesleytodd is suggesting.  Set a threshold so you know if there is a large drop from the last known good, but allow the threshold to be changed if that drop is ok and understood.

docs/drafts/ci-cd-guidelines.md
bnb: I observed this practice in multiple modules, which is why I added it. Happy to tweak this to optional/remove it if y'all want 👍 

Ref for where I saw it: 
- **Babel:** https://github.com/babel/babel/blob/master/.circleci/config.yml#L55-L56
- **Express:** https://github.com/expressjs/express/blob/master/.travis.yml#L23-L25
- **Webpack:** https://github.com/webpack/webpack/blob/master/.travis.yml#L10-L14
- **Commander:** https://github.com/tj/commander.js/blob/master/.travis.yml#L7-L9
- **React:** https://github.com/facebook/react/blob/master/appveyor.yml#L39-L40

docs/drafts/ci-cd-guidelines.md
bnb: Happy to make this change if there is no -1s to it 👍

docs/drafts/ci-cd-guidelines.md
bnb: Trying to provide some guidance for those who many not be familiar. There are two reasons I included this and worded how I did:

First, I specifically pulled this from my personal experience with TravisCI's offered Linux distros, which are defined as their codenames: 

- https://docs.travis-ci.com/user/reference/precise/
- https://docs.travis-ci.com/user/reference/xenial/
- https://docs.travis-ci.com/user/reference/trusty/

Second, in my personal experience, most _modern_ Linux servers end up being deployed with Ubuntu (with the exception of Docker containers). Part of the reason I phrased this how I did was an attempt to suggest using the distro with the largest community adoption.

All that said, happy to remove this if everyone agrees.

docs/drafts/ci-cd-guidelines.md
bnb: Agreed, happy to remove this if others agree 👍

docs/drafts/ci-cd-guidelines.md
bnb: Would y'all like me to change/remove this?

docs/drafts/ci-cd-guidelines.md
ljharb: It’s certainly a fine thing to do, but *not* doing it just means CI jobs take slightly longer, so i feel like it’s less of a best practice and more of an optional optimization.

ljharb:
One thing that’s very important, but not enough packages do, is running `npm ls` in CI to ensure their dependency graph is valid. I’d like to include that here.

bnb:
@ljharb totally fine with adding that. A couple questions:

1. Would you put this under testing or another category?
2. Is there an additionally needed step outside of `npm ls` to define that the tree is indeed returning what is expected?

ljharb:
I'd put it under testing, since it's ensuring that your package.json changes are valid.

No, after `npm install`, either `npm ls` exits zero, or your dep graph is invalid and you can't rely on anything working. This is particularly important in any ecosystem that uses peer deps, which includes babel, eslint, react, angular, vue, jest, gulp, grunt, etc.

docs/drafts/ci-cd-guidelines.md
bcoe: As @mhdawson says, the goal isn't to enforce that a pull request has coverage, the goal is to fail if we break the coverage machinery. A pull request should never drop coverage from `95%` to `55%`, this is most likely a change to Node.js that breaks the instrumentation mechanism.

bcoe: we sit at `95%` coverage today, if we set the threshold to `85%` there's no healthy scenario that would cause this significant of a drop.

docs/drafts/ci-cd-guidelines.md
wesleytodd: I think we should strive to provide users with help to make it easy to support as far back as possible. And then leave it up to them if they want to not follow our recommendations after that.  With a good ci setup it is easy to maintain backwards compact for most modules.  

IMO there are no downsides to setting out to have the largest range of support.  With the obvious note that we don't control any of these projects, so they are free to drop support for older versions if it suits them.

docs/drafts/ci-cd-guidelines.md
```diff
@@ -0,0 +1,38 @@
...
+
+## How?
+These guidelines are intended to be just that – guidelines. We will also include templates for common CI/CD providers and accept contributions of templates that implement these guidelines from the community.
+
+It's worth noting that projects **do not** need to use one of the provided templates, but **do** need to ensure that their CI/CD setup enables them to follow the required guidelines at the very minimum. Projects are by no means restricted to the required guidelines – more complex and dynamic configurations are more than welcome.
+
+## CI/CD Templates Should Include:
+
+- Tests
+  - Criteria (**Required**):
```
mhdawson: Can we actually say "required" since these are guidelines to help maintainers?  It is still good for them to implement 3 of 4 that might be "required".  Would it be better to say "Minimum best practice"

docs/drafts/ci-cd-guidelines.md
wesleytodd: I like this approach.  The hard part will be describing this to users in an actionable way as a "best practice".  @bcoe, do you have a particular process for picking a threshold which has worked for you in the past we could document here and recommend?

wesleytodd:
@ljharb Instead of `npm ls` couldn't we recommend just running `npm ci` which will ensure an install compatible with the `package-lock.json`?  Isn't this the same effect that it validates the install?

docs/drafts/ci-cd-guidelines.md
Eomm: Talking like a nooby as I am 😃 I agree that to give some practical suggestion could save time to the ones that are not familiar with this stuff and spend many hours to choose which one is better!

docs/drafts/ci-cd-guidelines.md
bcoe: @wesleytodd very much personal experience based; But I've noticed if you're fairly diligent about writing tests for any feature you add, without paying particular interest to test coverage, you'll usually land in the `89% - 93%` coverage range for a codebase.

Since improving coverage is something folks pay attention to on the Node.js project, I'd be pretty shocked if any one pull request ever dropped things below `90%` -- picked `85%` just to recommend even more of a buffer.

isaacs: Coverage is not a guarantee of quality, it's true.  However, reducing coverage is definitely a sign of reducing test quality.  Fail tests if coverage reduces.  If you aren't a fan of failing, don't reduce coverage.

I have been requiring 100% coverage of all my projects for the last year or so, and it's made me a better programmer.  It doesn't mean my code is bug-free, but at least I have an easier time tracking down bugs and making sure that my tests do what I think they do.

Asking for 100% right off the bat is probably unreasonable.  A ratchet mechanism like this is ideal.  Fail the test if it goes down, and it'll keep going up.

docs/drafts/ci-cd-guidelines.md
ljharb: The easy thing tho is to use whatever’s the default for CI (which will usually be Ubuntu)

ljharb:
@wesleytodd no, as it doesn’t validate anything. You can easily have a lockfile with invalid deps.

wesleytodd:
Didn't know it did not validate., good to know!  As for having a lock file on packages I get the point, I don't always follow the advice but feel like I should.  Maybe having a section on how to disable it is good as well.

mcollina:
I think we should have a paragraph on 2FA.

docs/drafts/ci-cd-guidelines.md
dominykas: Can this be worded as the consensus here? Which can be summarised with this pic: Coverage is not a quality measure per se, but it can be very beneficial to find untested code ([source](https://martinfowler.com/bliki/TestCoverage.html))

docs/drafts/ci-cd-guidelines.md
```diff
@@ -0,0 +1,38 @@
...
+  - Criteria (**Optional**):
+    - code coverage
+      - failure if code coverage is below a well-documented threshold
+- Caching
+  - Criteria (**Required**):
+    - At a minimum, cache `node_modules` (or equivalent) for your project
+- Operating System Versions
+  - Criteria (**Required**):
+    - Windows
+    - macOS
```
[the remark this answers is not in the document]
dominykas: I think considering only cloud scenarios is not enough - node and it's package ecosystem is heavily used for frontend and lots of developers work on Win/Mac, so it is very important for packages to behave well on the dev envs, not just on deployed instances.

That said, I do realize that maintaining compat is extra work (and hard work), esp on Win, esp if you only own a Mac. That is probably a broader topic to discuss for this group, as there's scope to help maintainers with that.

mcollina:
I think “flaky” CI is a significant burden to maintainers. I did this wrong several times, and it’s almosr ab unrecoverable situation. Helping out with getting a solid green would be extremely helpful.

docs/drafts/ci-cd-guidelines.md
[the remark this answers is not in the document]
ljharb: travis-ci has both mac and windows support, included in the free plan.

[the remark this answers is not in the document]
ljharb: i don’t agree that yarn matters in any way whatsoever since node doesn’t officially support yarn, while it does officially support macs and windows - the definition for what’s correct for yarn is “what npm does”, so if it’s tested on npm its already good enough for yarn.

However, the difference between OSs only matters to a tiny minority of packages, so I’d be in support of only mentioning Linux here for the common case.

docs/drafts/ci-cd-guidelines.md
chrkaatz: Wherever I worked in the past, may it be a big corporation or a company where the compliance standards are high, those companies prefer running their servers with rhel/centos based distros. When I wanted to run something on ubuntu I had to state out my reasoning and a business justification for my choice.

bnb:
@mcollina how would you like that to be encapsulated in a bullet point?

docs/drafts/ci-cd-guidelines.md
bnb: One clarifying question:

I'm realizing now that this is more of a requirement from the project. This document was originally intended to outline what needs to be implemented.

Is code coverage within that scope – is it something that needs to be a part of the CI/CD configs (like `.travis.yml`, `appveyor.yml`, etc.) or is this a project-by-project thing? If it is something that would be included in the templates or is this an additional/different section?

ljharb: While I'd say coverage is good for every project, the decision to use it imo should be made project by project.

docs/drafts/ci-cd-guidelines.md
bnb: If we're going to do that then we should probably document that outside the bullets rather than within them. Does that sound good to you/others?

docs/drafts/ci-cd-guidelines.md
ghinks: Although I'm a bit of a testing fan, I have just recently been reminded that some modules ( like the N-API conversion type ones for hardware etc) are not testable in the modern CICD sense. I am with @ljharb in that coverage is good. But in some cases like add on native code they are quite impossible without bespoke hardware.

Eomm:
> @ljharb Instead of `npm ls` couldn't we recommend running `npm ci` which will ensure an install compatible with the `package-lock.json`? Isn't this the same effect that it validates the install?

Thinking on the modules, the `package-lock.json` cannot be published in the npm registry, so could it be more user-friendly not run `npm ci` in CI/CD? So the module will be tested with the dep valid in that moment.

mcollina:
@bnb 

> @mcollina how would you like that to be encapsulated in a bullet point?

* The CI run should be green on repeated runs. Tests that fails intermittently are problematic, and a burden for maintainers.

---

@ljharb @Eomm 

I think that `package-lock.json` creates more harm than good in maintaining packages. I'm not [alone](https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514) in this opinion. It might be good to document this choice, with pros & cons, or just avoid mentioning it.

nschonni:
> could it be more user-friendly not run `npm ci` in CI/CD

The command is available in 5.7.1, so it might not work for everyones matrix, depending on what Node/NPM versions they're supporting. Think once 6 drops off LTS, all the latest LTS will meet this minimum requirement though

ljharb:
@nschonni yes, but supporting only LTS isn't necessarily a good thing, and as has been said, having a lockfile (needed for npm ci) is not necessarily a good thing for packages.

docs/drafts/ci-cd-guidelines.md
```diff
@@ -0,0 +1,38 @@
...
+
+## CI/CD Templates Should Include:
+
+- Tests
+  - Criteria (**Required**):
+    - `npm test`
+      - This can be any testing framework exposed via `npm test`.
+      - Many CI/CD systems run this implicitly for JavaScript projects. You do not need to explicitly define `npm test` in these cases.
+  - Criteria (**Optional**):
+    - code coverage
```
Eomm: Could we add also `vulnerability scanner` as optional criteria?
I saw in action today and I think could be a good alert for the maintainer

mhdawson:
@bnb I think the next step is an update to address remaining comments?

bnb:
@mhdawson yup! Will take a pass now. Apologies for the delay, been busy the last few weeks 👍

docs/drafts/ci-cd-guidelines.md
bnb: Since this is optional, should this be removed given @mcollina's comment?

docs/drafts/ci-cd-guidelines.md
bnb: The consensus here seems to be split – it's good in _some_ cases. Since this is in the _optional_ section, can it remain or should it be removed?

docs/drafts/ci-cd-guidelines.md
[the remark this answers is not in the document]
bnb: FWIW, I'd agree with @wesleytodd in expressing that we should support what Node.js supports, and help do the work to lower the barrier to entry for OSS projects to do the same.

My intent here is not to favor one CI provider over another – I've had personal experience only with TravisCI's Windows builds and had very few problems with them other than the Node.js version manager used by Travis failing once. I wasn't aware of the general state of poor support in the CI/CD ecosystem for Windows builds, which deeply saddens me since a [non-trivial amount of the downloads](https://nodejs.org/metrics/#operating-systems) from nodejs.org are for Windows.

For context on why this was an interesting and exciting inclusion to me, I started learning and eventually contributing to Node.js on a $200 PC that I purchased at WalMart. I've told multiple contributors and collaborators about the experience I had and often asked what I could to do help prevent others from having that experience.

That experience was horrid because Windows support was so dramatically lacking at the time. While things have gotten better both in Node.js core and in the ecosystem, I still always want to do whatever I can to help make sure the next person who uses Node.js on Windows doesn't have anything that resembles the same experience that I did when I started.

That said, I am going to remove myself from this specific discussion point since @mcollina suggested that it prioritizes one provider – the one I am currently employed by – over any other. That is expressly _not_ my intent. If y'all could seek consensus on this point, I (or anyone else with commit access) can update it accordingly 💖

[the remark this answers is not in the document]
vweevers: I do not share the view that the current text favors a single provider (which is not even mentioned) supposedly because "other alternatives are significantly worse". Personally I'm happy with AppVeyor (their build times have significantly improved) and Travis for Windows is promising. Before this thread, I never even considered Azure as an option.

That said, I agree that Windows should not be required (except for native addons). For most packages it indeed adds no value, provided that _other_ best practices are followed, e.g. use `path` when dealing with filenames. Could that be worth a separate guide?

mhdawson:
@bnb thanks :)

docs/drafts/ci-cd-guidelines.md
bcoe: I think that coverage is almost always good, as @dominykas puts it: "use it to find untested code in a module, not to prove code is good".

If someone is using a static typing tool (like TypeScript) or a smart linter (like standard) this can certainly solve some of the same variety of issues that you solve with test coverage -- I might group all of these types of tools under the category of optional, with a discussion of why they're potentially great indicators of a healthy codebase:

* find un-used variables.
* find incorrectly invoked function signatures.
* find code that just straight up has never been run.

bnb:
@mhdawson sorry for the delay! Had the changes done locally since my last comment but didn't commit + push. PTAL ✨

docs/drafts/ci-cd-guidelines.md
bnb: Updated this to reflect the suggestion here.

docs/drafts/ci-cd-guidelines.md
bnb: Updated the language per the suggestions here 👍

bnb:
Also, it is worth a reminder: this is not for the CI/CD of any individual project, which does seem to be what some of the comments have been geared toward – this is for generalized CI/CD templates to be included in this project as a resource for project maintainers to pull.

docs/drafts/ci-cd-guidelines.md
bnb: Removed this section.

docs/drafts/ci-cd-guidelines.md
bnb: I believe we should leave this as optional, given it's a guideline for templates to be submitted to be included in this repo rather than a list of requirements for projects to be included.

docs/drafts/ci-cd-guidelines.md
bnb: For what it's worth, I think this should remain as an **optional** criteria for CI/CD templates to be PRed into this repository, given that there seems to be an even split on the discussion – some find it valuable and some don't. Perfect fit for "optional" 😁

docs/drafts/ci-cd-guidelines.md
bnb: Done!

bnb:
@mcollina on the maintaining a green CI, I'm not sure how that actually fully fits into guidelines for CI/CD templates. What code could be in a CI/CD template (for example, `.travis.yml`) that could point to this?

That's more of a per-project criteria that is geared toward acceptance rather than CI/CD templates 👍

docs/drafts/ci-cd-guidelines.md
```diff
@@ -1,38 +1,54 @@
...
+**Optional** guidelines are, as the name indicates, optional for templates to include. They're _nice to have_ but by no means required.
+
+## CI/CD Guidelines
 
 - Tests
-  - Criteria (**Required**):
+  - **Required**:
     - `npm test`
       - This can be any testing framework exposed via `npm test`.
       - Many CI/CD systems run this implicitly for JavaScript projects. You do not need to explicitly define `npm test` in these cases.
```
ljharb: to clarify, your package.json needs to, but your CI config might not.

docs/drafts/ci-cd-guidelines.md
```diff
@@ -0,0 +1,54 @@
...
+## CI/CD Guidelines
+
+- Tests
+  - **Required**:
+    - `npm test`
+      - This can be any testing framework exposed via `npm test`.
+      - Many CI/CD systems run this implicitly for JavaScript projects. You do not need to explicitly define `npm test` in these cases.
+    - `npm ls`
+      - If `npm ls` exits with a `0` exit code, your build should pass
+      - If `npm ls` exits with a non-`0` exit code, your dependnecy graph is invalid and should fail the run
```
dominykas: Not sure about this one - in practice, I've had too many false positives (`peerDependencies` for sure, but possibly something related to `optionalDependencies` and lock files - can't recall) with this approach, TBH. `rm -rf node_modules; npm install` completing before `npm test` should be a good enough indicator of dep tree validity.

ljharb:
```suggestion
      - If `npm ls` exits with a non-`0` exit code, your dependency graph is invalid and should fail the run
```

@dominykas i'm quite sure on it; if `npm ls` exits nonzero, something is broken. if there's a false positive it's likely due to broken dependency declarations in one of your deps.

`npm install` completing does NOT indicate that your dep graph is valid in any way; it just indicates it's installable. That's what `npm ls` is *for*.

Eomm:
Hi @bnb, can I help you to address comments in order to complete this PR?

Eomm:
@ghinks I agree and then starting from this draft we can build CI/CD templates in order to verify all those issues that are not clear

mhdawson:
I agree with @Eomm and @ghinks lets land this as a draft and then we can continue to refine through PRs'.

PR #152, opened 2019-02-06
Wish to join the team

karansapolia:
I agree with the issues raised about old, widely used and unmaintained packages and the need for Node to ensure some security and support. I would like to contribute on issues and security maintenance of these packages and grow with the team and implement the effective solutions the community comes up with.


Eomm:
Hi @karansapolia , you are in the right place! ✌

Issue #142, opened 2019-01-29
[key packages] Pilot Packages


Eomm:
Split 1/3 of #105 

We can use to start collaborating with the authors to learn and iterate with.

Edit more info: 
to reach the [WG goals](https://github.com/nodejs/package-maintenance#goals) we need to start to help some pilot packages.
These modules will be helped by this WG adding support tools, documentation, fixes and all things to archive our goals and carry benefit to the package itself.
This WG will improve his guidelines, will learn different requirements and will be able to find patterns and solutions, test all the process and tools that are gonna to be built in order to evangelize the community and let the maintenance of an OSS package will be easier for maintainers.
What this WG ask to the pilot packages is:
+ explain the biggest problems and their priorities (as talk about #113 )
+ give feedback to our improvements/suggestion/PR

----

Express and all its direct and transient dependencies like [pillarjs](https://github.com/pillarjs) and [jshttp](https://github.com/jshttp) are the most mentioned package to start with.
@wesleytodd  has discussed a bit of this possibility on the last Express TC meeting so I think the Express team would be on board to participating.

## What to discuss here

+ We need to agree or not to start with Express as a pilot package. If "no" what else?
+ To define if we can and should add more "friendly" pilot packages

After that we could start to get "Directions of Help": the way package maintainers wish to receive help and define the first tasks.

[Ex:](https://github.com/nodejs/package-maintenance/issues/105#issuecomment-451286446)
Express needs help with repo cleanup, automation, documentation and, user support.

mhdawson:
Starting with Express sounds good to me, given that they are a key part of the ecosystem and it sounds like they are open to working with us.

wesleytodd:
While clearly I am a fan of working with Express, I think we need at least a few others to make sure we have a variety of input.  Anyone have other suggestions?

mcollina:
I think https://github.com/mqttjs/MQTT.js and dependencies are another good place to start.

mhdawson:
I agree we need to have several and MQTT sounds like a good one to add to the list.

Eomm:
I don't know other friendly packages that could collaborate with us in this pilot phase.

Appling manually the basic criteria we talk about, I think that [request](https://www.npmjs.com/package/request) could be a candidate: it is widely used and have a lot of issue and PR: it seems they need help to break all that work!

I looked [here](https://david-dm.org/), there is a nice bubble schema that show the libs per usage, I took the first one with many issue&PR opened. Nothing personal 😁

[...]

PR #162, opened 2019-02-12
Pilot packages - cold start

Eomm:
Hi,
based on the meeting of yesterday I send this PR for the issue #142 

The list is not completed, we are waiting for others feedback/comment/approvals, meanwhile, we could start with confirmed ones.

I was thinking if is it necessary to find out a "WG package-maintenance reference" per pilot package, not only for the module?

### DIFF
--- a/pilots/README.md
+++ b/pilots/README.md
@@ -34,6 +34,9 @@ Doing so we will create a wide range of tools in order to:
 The baseline to start with this work is:
 
 + Prepare a [survey](./SURVEY.md) to collect more informations as possible
++ Dispose the [output](#output) directory and send the survey with all the details to the maintainer reference
++ The Maintainer Reference should open a PR to this repository with the replies
++ Dispelling every doubt before merge the PR
 + Think out the feedback received in tasks we should do to help the package
 + Brief with other pilot packages feedbacks to find common needs
 + Define a roadmap
--- a/pilots/README.md
+++ b/pilots/README.md
@@ -1,0 +1,40 @@
+# Pilot Packages
+
+The pilot packages are the alpha tester of this working group. They will help us to archive our
+[WG goals](https://github.com/nodejs/package-maintenance#goals):
++ identifing the major problems a module have to face
++ defining priorities for our todo-list
++ verifing the benefits of the tools and process built
++ improving all the stuff created to evangelize the community
+
+The help provide to all these packages will be to think and to implement solutions for their problems,
+helping them to migrate to newer Node.js versions.
+
+Doing so we will create a wide range of tools in order to:
++ give to all the maintainers the possibility to use them
++ provide the resources to choose wisely the dependencies for business
+
+
+## List
+
+| Package* | Maintainer Reference |
+|----------|----------------------|
+| [Express][1]     | []()
+| [MQTT][2]        | [@mcollina](https://github.com/mcollina)
+
+
+
+[1] https://github.com/expressjs/express
+[2] https://github.com/mqttjs/MQTT.js
+
+*Ordered per joining time.
+
+## Approach
+
++ [Survey](./SURVEY.md)
++ Collect Feedback
++ Find patterns and use cases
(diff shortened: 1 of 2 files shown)


pilots/README.md
```diff
@@ -0,0 +1,58 @@
...
+[MQTT]: https://github.com/mqttjs/MQTT.js
+
+*Ordered per joining time.
+
+
+## Approach
+
+The baseline to start with this work is:
+
++ Prepare a [survey](./SURVEY.md) to collect more informations as possible
```
mcollina: How will the survey delivered? Should we send a PR?

Eomm: Yes, the maintainer could send the reply with a PR (maybe there will be the necessity of little clarifications). But of course, the approval will be more like a Q&A.

mcollina: This should be clarified here.

pilots/README.md
```diff
@@ -0,0 +1,58 @@
...
++ give to all the maintainers the possibility to use them
++ provide the resources to choose wisely the dependencies for business
++ improve the quality and stability of the packages in the Node.js ecosystem
+
+
+## List
+
+| Package* | Maintainer Reference |
+|----------|----------------------|
+| [Express]     | [@wesleytodd](https://github.com/wesleytodd)
```
wesleytodd: To be clear here, I think we should change this to Organization, or Ecosystem.  It is not the `express` package specifically that needs attention.  It is the entire ecosystem of modules that make up what people think of when they say "express".  This includes all the modules in [jshttp](https://github.com/jshttp) and [pillarjs](https://github.com/pillarjs) as well as the ones directly in the [express org](https://github.com/expressjs).

Eomm: Sorry if it is a silly question but, in this case, our reference will be anyway Express right?

pilots/README.md
Eomm: Done

pilots/README.md
wesleytodd: Well I am not sure there will be much to apply in the express repo proper.  My guess is most of the recommendations and changes will be in the other repos, which is why I mentioned it.

Eomm: Clear, I'm thinking only how to match this use-case with the "SURVEY.md", I think I should add some questions like:
"which problems are due your dependencies?"

wesleytodd: FWIW, this is the kind of thing I am referring to when I say "not express, but express repos".

https://github.com/pillarjs/hbs/issues/162

There are a bunch of modules which Doug is maintaining or in some way managing.  See all the repos in `pillarjs` and `jshttp` as mentioned above.  These are the ones the Express TC is responsible for, and are short staffed for.

mhdawson:
@wesleytodd, @mcollina look good to you, looks like your comments were addressed.

pilots/README.md
```diff
@@ -0,0 +1,61 @@
...
++ provide the resources to choose wisely the dependencies for business
++ improve the quality and stability of the packages in the Node.js ecosystem
+
+
+## List
+
+| Organization* | Maintainer Reference |
+|---------------|----------------------|
+| [Express]     | [@wesleytodd](https://github.com/wesleytodd)
+| [MQTT]        | [@mcollina](https://github.com/mcollina)
```
mcollina:
```suggestion
| [MQTT.js]        | [@mcollina](https://github.com/mcollina)
```

pilots/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+[Express]: https://github.com/expressjs/express
+[MQTT]: https://github.com/mqttjs/MQTT.js
```
mcollina:
```suggestion
[MQTT.js]: https://github.com/mqttjs/MQTT.js
```

mhdawson:
@Eomm if you can accept the suggested  changes we can get this landed.

Eomm:
@mhdawson done 👍

Issue #157, opened 2019-02-11
[baseline practices] Testing


rxmarbles:
This is in continuation of #119 to keep the discussion focused on one part as we can further flesh out each piece to create a draft proposal around baseline practices. 

# Testing
  * Jest, Mocha
  * coverage should match most use cases for the module

_If I am missing anything please add a comment and tag me so I can update_

ljharb:
Also https://npmjs.com/tape, which is the only test runner I'm aware of that works on older nodes - my typical advice is to use jest or mocha for apps, and tape for packages, since packages should be supporting as old an engine as possible (#156).

mcollina:
I'm personally a huge fan of tap. I found that mocha tests tends to be brittle and flaky, because it runs all files within a single process. If cleanup is not done correctly, it will be really hard to prevent bad interactions between the tests.

wesleytodd:
> Overall I do not think we should be expressing preferences here

Completely agree!  I have had the opposite experience with Mocha, it has been more stable than other runners because of its single process methodology :).  The take away from this is that we should be agnostic to technical implementation.

rxmarbles:
agree on not expressing preferences and the proposal should focus on the available testing frameworks/methodologies.

mhdawson:
I agree we should provide recommendations on "what" should be done, and then add information on options for "how" to do that.

ahmadnassri:
should start with a simple context of "Why", assuming lowest level of knowledge, starting with alignment on the value of testing, why it's needed, and what benefits it provides to both the package maintainer and the users.

craftninja:
From discussions in meeting April 22 #196 

- [ ] guidance around "how do I go from 0 to some testing"
- [ ] "there should be something in a module's package so you run `npm test` and see output, very clear whether you pass or fail"

dominykas:
Note to self: extend both - testing and ci/cd guidelines with additional info (and link to) the "Please test in LTS" blogpost (#260).

boneskull:
A testing guidelines doc has been merged, written by @ghinks.  Should this issue be closed?

mhdawson:
closing

PR #169, opened 2019-03-01
A first draft of testing guidelines

craftninja:
Maybe changes:
- [x] ~links to good examples~ not yet
- [x] ~discuss move to ci-cd guidelines? or does some of that content move here? how do we differentiate between these to guideline pages?~ we will have both sections for now

### What is this?
Hello friends!

This is a first draft of testing guidelines. I started using the conversation in #157

### DIFF
--- a/docs/drafts/testing-guidelines.md
+++ b/docs/drafts/testing-guidelines.md
@@ -18,6 +18,7 @@ Here are some things to consider as you develop your package.
 
 Some useful tools:
 * [c8](https://www.npmjs.com/package/c8)
+* [citgm](https://www.npmjs.com/package/citgm)
 * [Codecov](https://www.npmjs.com/package/codecov)
 * [Coveralls](https://www.npmjs.com/package/coveralls)
 * [Jest](https://www.npmjs.com/package/jest)
--- a/docs/drafts/testing-guidelines.md
+++ b/docs/drafts/testing-guidelines.md
@@ -17,6 +17,11 @@ It is a good idea to have unit tests, coverage that matches most use cases for t
 Here are some things to consider as you develop your package.
 
 Some useful tools:
+* [c8](https://www.npmjs.com/package/c8)
+* [Codecov](https://www.npmjs.com/package/codecov)
+* [Coveralls](https://www.npmjs.com/package/coveralls)
 * [Jest](https://www.npmjs.com/package/jest)
 * [Mocha](https://www.npmjs.com/package/mocha)
+* [Nock](https://www.npmjs.com/package/nock)
+* [nyc](https://www.npmjs.com/package/nyc)
 * [tape](https://www.npmjs.com/package/tape)
--- a/docs/drafts/testing-guidelines.md
+++ b/docs/drafts/testing-guidelines.md
@@ -8,7 +8,11 @@ Testing is useful for developing new features, refactoring with confidence, and 
 Testing is even more important when new maintainers take over a package. Sometimes when we work on a codebase for a long time we forget to articulate all the cognitive load to which we have become accustomed. Good test coverage can alleviate this burden.
 
 ### How?
-It is a good idea to have unit tests and coverage that matches most use cases for the module.
+It is a good idea to have unit tests, coverage that matches most use cases for the module, and make sure things work in all supported environments.
+
+* **unit test**: test your code
+* **integration test**: test your code with other applications dependencies
+* **acceptance test**: test your application sticks in performance, heavy load, etc..
 
 Here are some things to consider as you develop your package.
 
--- a/docs/drafts/testing-guidelines.md
+++ b/docs/drafts/testing-guidelines.md
@@ -1,0 +1,18 @@
+# Testing
+
+### Why?
+Testing is awesome!
+
+Testing is useful for developing new features, refactoring with confidence, and making sure new things don't break old things. When other people rely on your code for their own applications, testing helps you make sure things don't break for more than just yourself!
+
+Testing is even more important when new maintainers take over a package. Sometimes when we work on a codebase for a long time we forget to articulate all the cognitive load to which we have become accustomed. Good test coverage can alleviate this burden.
+
+### How?
+It is a good idea to have unit tests and coverage that matches most use cases for the module.
+
+Here are some things to consider as you develop your package.
+
+Some useful tools:
+* [Jest](https://www.npmjs.com/package/jest)
+* [Mocha](https://www.npmjs.com/package/mocha)
+* [tape](https://www.npmjs.com/package/tape)


docs/drafts/testing-guidelines.md
```diff
@@ -0,0 +1,18 @@
...
+Testing is useful for developing new features, refactoring with confidence, and making sure new things don't break old things. When other people rely on your code for their own applications, testing helps you make sure things don't break for more than just yourself!
+
+Testing is even more important when new maintainers take over a package. Sometimes when we work on a codebase for a long time we forget to articulate all the cognitive load to which we have become accustomed. Good test coverage can alleviate this burden.
+
+### How?
+It is a good idea to have unit tests and coverage that matches most use cases for the module.
+
+Here are some things to consider as you develop your package.
+
+Some useful tools:
```
ljharb: It’d be good to clarify browser/engine support of each, paired with a statement higher up that it’s important to test in all supported environments

docs/drafts/testing-guidelines.md
```diff
@@ -0,0 +1,18 @@
+# Testing
+
+### Why?
+Testing is awesome!
+
+Testing is useful for developing new features, refactoring with confidence, and making sure new things don't break old things. When other people rely on your code for their own applications, testing helps you make sure things don't break for more than just yourself!
+
+Testing is even more important when new maintainers take over a package. Sometimes when we work on a codebase for a long time we forget to articulate all the cognitive load to which we have become accustomed. Good test coverage can alleviate this burden.
+
```
Eomm: Could be useful to add also a line for:
+ unit test: test your code
+ integration test: test your code with other applications dependencies
+ acceptance test: test your application sticks in performance, heavy load, etc..

ljharb: I’d be careful about unit vs integration; unit tests can also integrate across all dependencies.

craftninja: @Eomm Added your bullet points to the `How` section, see if that flow looks good for you.
@ljharb - Do you think something needs to be reworded or added to reflect this concern?

docs/drafts/testing-guidelines.md
craftninja: @ljharb I added something about making sure things work in all supported environments, do you think that is clear enough? Or do we want to go into details about how?

docs/drafts/testing-guidelines.md
ljharb: I would avoid categorizing types of tests at all.

Eomm: @ljharb When I said "dependancies" I was meaning a DB or whatever outside of the module domain. I think that this concept let programmers design better their module because if they write all global vars (for example) they will have pain also to write some unit-test. For this example, I think that there is still confusion (for what I see in my daily job) and focus on "target" words could unlock a new design-thinking. Of course, a BIG package doesn't need this information, but when I try to figure out who will read the docs we are writing, I'm thinking of a young dev that wants to write the "perfect module" and I would like to know that exists multiple types of test so I can go deeper ✌
What do you think of this vision?

ljharb: I think that it’s such a complex subject that we shouldn’t delve into it.

What you mean isn’t “dependencies” (which in this context are the things npm install installs) but low-level i/o - the network, the filesystem, etc. Certainly your tests can either mock these out or exercise them. However, the common case is to mock out low-level i/o and otherwise unit test (by which i mean, fully integration test all of your code, except for that low-level i/o) all of your code’s observable API and semantics.

mhdawson:
It would be great if we could include advice that would help the tests be run in tools like CITGM.  This might include using a common target and ensuring that the result is easy to identify as passed or failed.

mhdawson:
Some info on how to generate/track coverage might also be useful  (@bcoe FYI)

mhdawson:
Agree versus the potential overalp with the CI/CD we might start with cross/references but I do think it is important that we have a 'template' whether that be suggestions or templates that can be re-used.  Maybe even a script which adds the recommended target to the package.json and any other concreate artifacts that can be generated.  The closer we can get to adding what's needed (in the simple case) versus recommeding what to do the better.

docs/drafts/testing-guidelines.md
bnb: @ljharb could you clarify what you mean? I think I have an idea but I'm pretty sure I'm incorrect.

docs/drafts/testing-guidelines.md
```diff
@@ -0,0 +1,18 @@
...
+### How?
+It is a good idea to have unit tests and coverage that matches most use cases for the module.
+
+Here are some things to consider as you develop your package.
+
+Some useful tools:
+* [Jest](https://www.npmjs.com/package/jest)
+* [Mocha](https://www.npmjs.com/package/mocha)
+* [tape](https://www.npmjs.com/package/tape)
```
bnb: Perhaps add these, since they're useful tools (but _aren't_ necessarily testing libs):
```
* [c8](https://www.npmjs.com/package/c8)
* [nyc](https://www.npmjs.com/package/nyc)
* [coveralls](https://www.npmjs.com/package/coveralls)
* [codecov](https://www.npmjs.com/package/codecov)
* [nock](https://www.npmjs.com/package/nock)
```

docs/drafts/testing-guidelines.md
bnb: I tend to agree with @ljharb here. If needed, this can be expanded later in a different proposal.

docs/drafts/testing-guidelines.md
ljharb: What i mean is, most test runners don’t support older nodes; jest and mocha require 6+, etc. Before choosing a tool, people should be informed as to the support it offers, and ide.ly should be encouraged to choose the tools that allow the widest support possible.

docs/drafts/testing-guidelines.md
craftninja: Ordering in "importance" seems like it could be a long discussion, so added, and alphabetized, and cased according to documentation / descriptions from npm.

Updated!

docs/drafts/testing-guidelines.md
craftninja: So do we want to delete contents in `### How` section but leave `Some useful tools:` section?

@mhdawson recommended adding some advice around tools like citgm. I might wait and see if we want to pull back from definitions and recommendations for now... but be glad to add/remove if we think we need *some* or *both* or *none*.

docs/drafts/testing-guidelines.md
craftninja: @mhdawson Do you think these updated links are "enough" for now regarding citgm and tracking test coverage?

bcoe:
I'd be happy to write up a blurb about why I *love* test coverage. Is this something you'd be open to @craftninja; could be a follow up PR sometime soon.

craftninja:
@bcoe For sure, I would definitely be open to this. I'd be glad to add you to my fork for this PR as a contributor, or feel free to try out the new "suggestion" feature if that works (not sure if you end up on the diff or not 🤔 ).

Screen Shot 2019-03-25 at 2 16 04 PM

mhdawson:
Removing from the agenda, please put back on if/when you want it to be discussed in the next meeting.

Issue #160, opened 2019-02-11
[baseline practices] Choosing a License and Choosing Dependencies


rxmarbles:
This is in continuation of #119 to keep the discussion focused on one part as we can further flesh out each piece to create a draft proposal around baseline practices. 

# Choosing a License
  * important to the end user of which license the package uses

# Choosing Dependencies
  * has there been a change of ownership ? re-review dep : nothing

_If I am missing anything please add a comment and tag me so I can update_

ahmadnassri:
https://choosealicense.com/ is a great reference for newcomers into the world of OSS Licensing.

mhdawson:
@rxmarbles will take a first pass.

ghinks:
Glad to lend a hand as I have been meaning to understand all the options we are given in the github LICENSE file, been going for the unlicensed recently!

ljharb:
We should certainly mandate choosing *some* license, and ideally discourage "unlicensed".

ahmadnassri:
> We should certainly mandate choosing _some_ license, and ideally discourage "unlicensed".

I would agree, with emphasis on the context of Open Source, however, many _-if not all-_ of the guides written here willl also apply to inner-source projects, and I would ideally like to see emphasis on things like "SHOULD" / "MUST" be presented in the right context...

mcollina:
Note that we should not suggesting to relicensing. A license should be a MUST.

We _might_ decide that GPL and AGPL code is not part of this program, as well as "patent" licenses.

ljharb:
Yes, it’d be great if we could also discourage viral licenses.

vweevers:
Could you explain why those licenses are at odds with this project?

ljharb:
They discourage a healthy ecosystem - a virally licensed product can not be used by most companies, and as such, either the package, or most companies, are excluded from participating in that part of the ecosystem.

vweevers:
I hope node can be inclusive to both groups. Companies are not saints and copyleft is not a cure-all; both groups have valid viewpoints. Node could even play a role in removing some of the fear surrounding copyleft code and using it in a company (reminder: pretty much everyone is using the GPL-licensed Linux kernel, one way or another). This could bring the groups closer together. Discouraging copyleft achieves the opposite.

ljharb:
There’s a bit of a difference between licenses that allow versus licenses that constrain; the fear you refer to comes from most companies’ legal teams, and short of a precedent-setting court case, that’s unlikely to change.

[...]

PR #173, opened 2019-03-09
add basic wording for licensing and open the debate up to PRs.

ghinks:
# [Issue 160 licensing](https://github.com/nodejs/package-maintenance/issues/160)

In order to get the ball rolling with some actual prose. This may serve as a starting point. I'm not making any recommendations either for or against individual licenses. Rather I'm saying that we 
don't want
- restrictive
- patent

and do want
- the owner to have a license
- choose one appropriately 

I think it is not likely we actually would come across a project we want to support that has no license.

### DIFF
--- a/docs/drafts/licensing.md
+++ b/docs/drafts/licensing.md
@@ -1,0 +1,11 @@
+# Licensing
+
+It is required that any package supported have an appropriate license. The 
+maintainers may choose not to support a package with a license that is restrictive 
+or may be construed as a patent license.
+
+There are many licenses to choose from. We would refer package owners who have not 
+already choosen a license to review the [commonly used SPDX licenses](https://spdx.org/licenses/) and 
+[suggested means to choose license](https://choosealicense.com/).
+
+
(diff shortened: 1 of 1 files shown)


ljharb:
Could we determine the most used licenses on npm (by querying the license field in package.jsons in all packages), and report those statistics here? That would serve as a recommendation without making anyone worried that we were giving legal advice.

ghinks:
I think that is an excellent idea. Let me see what data the NPM folks have on that. I have an owed favor.

ljharb:
@ghinks if they can provide it readily, awesome - otherwise we can use @ChALkeR’s tool to just look at al the public packages.

@Eomm excellent idea; it should be present and valid SPDX.

dominykas:
Do we recommend/discourage using a LICENSE file?

ljharb:
Recommend, alongside the SPDX field. Both are required.

ghinks:
( in order of popularity )
- MIT
- Apache
- BSD
- ISC
- SEE LICENSE
- MPL
- [object Object]
- GPL
- CC0-1.0
- LGPL
- Unlicense
- pemrouz.mit-license.org
- Artistic-2.0
- WTFPL
- AGPL

1. The license is literally "SEE LICENSE", i.e. it refers to a license.txt
2. [object Object] is again literally the license field when people put JS in it
3. "Unlicense" is "The Unlicense", not un-licensed code
4. The WTFPL really is that popular

ljharb:
to clarify; `[object Object]` is when they use the antiquated object form of the field, which still, however, can be normalized to a license - unfortunately they didn't seem to apply that normalization.

PR #180, opened 2019-03-21
add myself as a new member

fharper:
I would love to participate in this working group on a personal level, but also as part of working at npm, which will give me the opportunity to use our internal expertise if/when needed.


fharper:
thanks!

PR #184, opened 2019-03-25
+ Survey: version question

tomByrer:
Usually use a [VSCode plugin](https://marketplace.visualstudio.com/items?itemName=pflannery.vscode-versionlens) to see if there are new versions in my package.json.  Unfortunately, it doesn't list `pre-release` or `next` versions which can [hide needed fixes](https://github.com/videojs/video.js/issues/5771).

Also, repos are scattered how they summarize changes.  Some do `CHANGELOG.md`, some write long descriptions in their release post, some just list the PR titles, & a few still use `README` or blogs.

### DIFF
--- a/pilots/SURVEY.md
+++ b/pilots/SURVEY.md
@@ -12,6 +12,7 @@ These are the basic questions to start with:
 + What are the most frequent problems?
 + What would you like to do to us?
 + What problems are due to your dependencies?
++ How do you discover if a new version of your dependencies are available and what changes are made?
 + Are your dependencies need of help? 
 + Do you have to contribute to your dependencies? Do you establish a good communication channel?
 


pilots/SURVEY.md
```diff
@@ -12,6 +12,7 @@ These are the basic questions to start with:
 + What are the most frequent problems?
 + What would you like to do to us?
 + What problems are due to your dependencies?
++ How do you discover if a new version of your dependencies are available and what changes are made?
```
Eomm: Does this question aim to find a pattern in order to automate those checks?

ljharb: typically one uses a service like greenkeeper or renovate for this.

tomByrer: @ljharb I've only seen a percentage of projects use a bot like Greenkeeper to update deps.  Many projects don't even have CI set up.

The question could find who uses update-bots, or who updates manually or like me semi-manually.  I'll often check release notes before updating major deps.  Sometimes the npm version is not new enough for me.

Eomm: This survey is for widely used modules, so I assume they have this kind of stuff.

I'm not pro nether contrary to this addition

mhdawson: I don't think it hurts to ask as we might find that our assumptions are incorrect.

PR #203, opened 2019-05-06
Actions for pilot package MQTT.js

Eomm:
Following [this](https://github.com/nodejs/package-maintenance/tree/master/pilots) output, I'm opening this PR to discuss on the possible tasks we could do to help the MQTT.js pilot package.

This is a raw list but I think it could be smoother building useful gadget that could be reused of other modules.


mhdawson:
Related I created this project where we can start opening issues that we want people to help out with.  https://github.com/nodejs/package-maintenance/projects/2.

PR #255, opened 2019-09-10
Rename "best-effort" to "ad-hoc"

kemitchell:
I noticed the current draft of the support format uses "best effort", a problematic legalism common in contracts that's usually taken to mean rather the opposite of what the group seems to mean here. I recommend replacing "best effort" with something else, and propose "ad hoc" as a starting point. I also considered "impromptu", "informal", "sporadic", and "casual". I'd rather avoid Latin, but couldn't find an earthy Anglo-Saxon synonym.

"Best efforts" and the like are strongly discouraged usages in legal drafting. See, for example, [a post from Ken Adams](https://www.adamsdrafting.com/what-does-best-efforts-mean/), a well known drafting specialist.

When lawyers use them anyway, the question often becomes how much pain, inconvenience, and expense the party that promised best efforts has to endure to get the job done. Since "best" is a superlative, the expectation is that in nearly every normal case, the job will get done. Not that the job will get done if one feels like it, has time, takes an interest, and so on.

I have a broader, related concern about the way that a standard like this could put pressure on already stretched maintainers to make essentially gratuitous support SLA commitment to all comers. But as much as I care about that more than this word-choice issue, I'd prefer to table the broader issue for now. It's more than I can bite off right now.

### DIFF
--- a/docs/drafts/PACKAGE-SUPPORT.md
+++ b/docs/drafts/PACKAGE-SUPPORT.md
@@ -196,7 +196,7 @@ The standardized support identifiers are as follows:
 | Value | Description |
 |-------|-------------|
 | `none`         | Don't expect a response, the package is not being actively maintained
-| `ad-hoc`       | The maintainer is interested in fixing/discussing issues, however, there should be no expectation on response times. If and when the maintainer has time they may respond.
+| `ad-hoc`       | The maintainer is interested in fixing/discussing issues. However, there should be no expectation on response times. If and when the maintainer has time, they may respond.
 | `regular-X`    | There are dedicated resources who regularly maintain the package, expected response time is X days or less for a "we read your issue" response. Further work will depend on prioritization of the issue by the maintainer team.
 | `24-7`         | There are dedicated resources who regularly maintain the package and they are available 24/7. You can expect to be able to contact the maintainers and get an initial response with 6 hours.
 
--- a/docs/drafts/PACKAGE-SUPPORT.md
+++ b/docs/drafts/PACKAGE-SUPPORT.md
@@ -60,7 +60,7 @@ The default for packages created by individuals for their own use should be:
           "node": "none"
         },
         "response": {
-          "type": "best-effort",
+          "type": "ad-hoc",
           "paid": false,
           "contact": {
             "name": "Volunteers",
@@ -153,7 +153,7 @@ For example:
 ```json
 response: [
   {
-    "type": "best-effort",
+    "type": "ad-hoc",
     "paid": false,
     "contact": {
       "name": "Volunteers",
@@ -196,7 +196,7 @@ The standardized support identifiers are as follows:
 | Value | Description |
 |-------|-------------|
 | `none`         | Don't expect a response, the package is not being actively maintained
-| `best-effort`  | The maintainer is interested in fixing/discussing issues, however, there should be no expectation on response times. If and when the maintainer has time they may respond.
+| `ad-hoc`       | The maintainer is interested in fixing/discussing issues, however, there should be no expectation on response times. If and when the maintainer has time they may respond.
 | `regular-X`    | There are dedicated resources who regularly maintain the package, expected response time is X days or less for a "we read your issue" response. Further work will depend on prioritization of the issue by the maintainer team.
 | `24-7`         | There are dedicated resources who regularly maintain the package and they are available 24/7. You can expect to be able to contact the maintainers and get an initial response with 6 hours.
 


docs/drafts/PACKAGE-SUPPORT.md
```diff
@@ -196,7 +196,7 @@ The standardized support identifiers are as follows:
 | Value | Description |
 |-------|-------------|
 | `none`         | Don't expect a response, the package is not being actively maintained
-| `best-effort`  | The maintainer is interested in fixing/discussing issues, however, there should be no expectation on response times. If and when the maintainer has time they may respond.
+| `ad-hoc`       | The maintainer is interested in fixing/discussing issues, however, there should be no expectation on response times. If and when the maintainer has time they may respond.
```
Trott: Totally optional nit-pick, but since we're in here anyway, maybe fix up some problems with the text?

```suggestion
| `ad-hoc`       | The maintainer is interested in fixing/discussing issues. However, there should be no expectation on response times. If and when the maintainer has time, they may respond.
```

kemitchell: @Trott took this since it was easy, but I'd prefer to keep this PR focused overall.

I'm just dipping a toe here. My availability is really limited right now.

kemitchell:
Thanks, @ljharb. I agree that once we step away from the "best effort" landmine, this becomes a bikeshed.

wesleytodd:
I agree we should be careful to protect authors as much as possible.  Having a lawyer view here is incredibly valuable.  Would `as-available` do?  

I don't really have a problem with `ad-hoc`, but it seems a bit more obscure in meaning than some of the other tags.  I know we are being very "english-centric" here, but I assume that "plain english" is better than specialized terminology for international understanding?

Also, `ad-hoc` to me can imply even less support as sometimes the term is used to mean temporary or to use one of your other proposed terms sporadic.  I am not sporadic in my response, but I think most of my projects receive "as available" support, and might not fit the legal definition of "best effort" but that is analogous in my mind.

kemitchell:
Maybe something like `no-commitment` or `no-sla`?

We're bikshedding.

lholmquist:
To add to the bike-shed,  what about just `PRs-welcome` :)

wesleytodd:
For sure bikeshedding, sorry that I started it but I think if we are going to change it doesn't hurt to do it a bit now 😄 

> no-commitment or no-sla

I am trying to capture the meaning of "I will try hard to help, but if I am on vacation or busy with other things, this might slip".  These to me sound more like "no support given", which sends the wrong message IMO.

ljharb:
"good-intentions"

kemitchell:
`try-hard`

vweevers:
`willing`
`open`

mhdawson:
how about `time-permitting`

Eomm:
Emoji vs bikshedding

in order of appearance:

* ad-hoc --> put an approve (sorry there is only 8 emoji for comments)
* as-available 👍
* no-commitment 👎
* no-sla 😄
* good-intentions 🎉
* try-hard 😕
* willing ❤
* open 🚀
* time-permitting 👀

Vote!

kemitchell:
I'm sorry, fellas. It's not fair to anyone else pretending I have spare cycles.

I'm going to go ahead and close the PR.  Please feel free to consider it like an issue instead, expressing concern with `best-effort`.

Eomm:
I'm sorry for this, my intention was to reach consensus regarding the word that best fits your great analysis and clear meaning for all the community

As said https://github.com/nodejs/package-maintenance/pull/255#issuecomment-530025663, for italians `ad-hoc` means something that is special, like a party or an exception.

> Please feel free to consider it like an issue instead

It worth 👍

kemitchell:
Nothing to apologize for, @eomm.

wesleytodd:
I am 👍 for reopening this and letting the conversation land on a better therm, then just updating the PR to reflect that consensus.

kemitchell:
@wesleytodd I don't have bandwidth to keep pushing this PR. Feel free to open one of your own.

wesleytodd:
Looks like we have a runoff between `as-available` and `time-permitting`.  Anyone have an opinion not already expressed?

lholmquist:
i'd vote for `time-permitting`

mhdawson:
Just catching up and added my vote for "time-permitting" if that helps

mhdawson:
Michael to take to Foundation to see if we can get any advice to see to see if any one is better than the others.

mhdawson:
Request that we narrow down to the choice we'd prefer so that we can ask a more targetted question:

```
I can certainly do so, but I'd also note that this isn't a term in a legal document. There's always a balance between asking attorneys for answers and incurring billable hours, as well...

With this in mind, I might suggest first narrowing this down to the choice that makes the most sense in the context? If there are still outstanding concerns at that point, I can help formulate a specific, scoped question so that we get back a specific, scoped answer.
```

mhdawson:
From my understanding of the response in https://github.com/openjs-foundation/cross-project-council/issues/335#issuecomment-538049573 is that the advice is we should be comfortable choosing the one we think fits best.

mhdawson:
Any strong objections to going with `time-permitting` which seems to have the most support?

mhdawson:
@kemitchell discussion in package maintenance meeting, and in issue.  Consensus seems to be time-permitting.  Can you update to that and then we'll land.

PR #258, opened 2019-09-15
Add @gntem to members.md

gntem:
Working with Node.js & NPM daily would like to participate more actively in this group.

✔️ Would love to help with any maintenance work needed to be done.


Eomm:
See you at the meeting and/or in the issues!
We have applied a new label if you want to start digging into it 👍

PR #267, opened 2019-10-12
best-effort to time-permitting

Eomm:
Follow up #255


mhdawson:
I think we've discussed this extensively in other issues, have a good number of approvals and it's been open >48hours so landing.

PR #276, opened 2019-10-30
Add @helio-frota as member

helio-frota:
Hi, 
I'm interested in becoming a member. We are continuing to explore our "supporting" options for modules in the Node ecosystem here at Red Hat also I'd like to get involved with [Express contributions triage efforts](https://github.com/expressjs/discussions/issues/92).


wesleytodd:
Awesome @helio-frota!  Welcome, your help will be much appreciated.  To get started it would be great if you could comment over there so that we can build momentum with the TC about getting this going.

Issue #244, opened 2019-08-27
CI Publish with Two-Factor Auth for Teams


wesleytodd:
All package publishes should be done with 2FA enabled. This requirements means that publishing from CI is difficult because it requires the CI to have a OTP from the user.  In order to maintain the 2FA you need some third party setup (a api to orchestrate the 2FA).  @dominykas and the team at Near Form have done some exploratory work on this with [Optic](https://github.com/nearform/optic/).  This situation is even more complicated for teams managing publish access (like we want for Express).

The POC solves well the CI portion for a single maintainer, but it requires a infrastructure which I am not sure is best long term for OSS projects. It uses web push, so requires a Firebase account.  Any infrastructure requirement is complicated for typical OSS maintainers. This is also vendor lock in.

In looking for options for a setup in Express, @dominykas and I hoped on a call to discuss.  The outcome from that call was that we identified one clear best case scenario:

Npm builds a release manager which supports 2FA.  So the idea would be that when CI pushed a release, it would be held in a pending state if it requires 2FA but no OTP was provided. You could then visit the website to see a list of pending releases. There you provide your 2FA OTP upon "approving" the release.

I am sure there are design and technical considerations to this approach, but we thought we should bring it up here and see what people think about this proposed solution.  There are other approaches we discussed, and if I have time I will follow up with descriptions of those, but I wanted to get this posted to open up the conversation asap.

Eomm:
I think 2FA it is a matter of security where you have doubled the things a thief must steal.
So if you maintain these two things separated you could think to avoid the "dead man's switch".
Anyway, I like to have control of the publish, but maybe I push the publish button not so often.

[fun note] I built up some workarounds for this purpose with a bunch of server in different countries a many firewall (I was constrained 🤣):

```
const twoFactor = require('node-2fa')
var newOtp = twoFactor.generateToken('MY BOSS TOTP TOKEN')
console.log( { newOtp } ); // use the OTP in your CI publish
```

mhdawson:
@darcyclarke @ahmadnassri, has there been any discussion within npm on this topic that you could share?

wesleytodd:
I heard on the twitters that some sort of staged release was on the roadmap at one point 🤞

dominykas:
@dominykas to write up things we can do with and without npm's involvement.

darcyclarke:
@mhdawson @wesleytodd Apologize for the delayed response here.

I'll see if I can't address a few of the points made in the initial issue:
[...]

PR #282, opened 2019-11-10
2FA in CI

dominykas:
The primary intent of this PR is to kick off a discussion on the various options around _how_ 2FA could be used for publishing from CI, incl. by teams.

I listed multiple options, I'd like to hear feedback about them (I might have missed some pros/cons) and it would be nice to hear some other ideas (feel free to push commits with them, if you have to).

If we can get an agreement on which ones are worth pursuing the most - we could possibly extract them into separate issues/projects either here or under `@pkgjs`, but for now I just wanted to have everything in one place as a starting point.

I know there's opinions about whether it is a good idea to publish from CI, but I think that's a discussion for CI/CD and publishing guidelines - the fact is that [people _are_ publishing from CI](https://blog.greenkeeper.io/introduction-to-semantic-release-33f73b117c8) and they are _not_ using 2FA, and that needs to be fixed (and I don't think that asking them to not use CI for publishing is something that would bring any results).

See also: #244 

---

As a personal note, after I wrote down the "poor man's version" of the Release Manager approach (i.e. setting the 2FA flag on a published release post-factum), I've kind of grown to think that it might just be one of the more awesome ways forward.

At the same time, I'm actually quite excited about all the options, and in an ideal world all of them would be available for people to use.


docs/drafts/2fa.md
```diff
@@ -0,0 +1,165 @@
+# Baseline practice: using 2FA when publishing
+
+To avoid malicious use of stolen package registry tokens, it is advisable to enable two-factor authentication (2FA). When it is enabled every call to `npm publish` requires a unique one time password (OTP) to be entered.
+
+To make sure that the code is a _second_ factor it should be generated by a human on an independent device. The secret seed of the OTP generator should be protected in such a way, that it cannot be obtained together with the registry token.
+
+At the moment, the advice to use 2FA is at odds with advice to automate the testing and release processes, esp. with publishing automatically from Continuous Integration (CI) systems (e.g. Travis).
```
vweevers: Where is the advice that the release process should be automated? Opinions differ.

ljharb: I very strongly believe that publishes - the only irrevocable thing - should never be automated, for many reasons (mostly around safety and auditing and having humans as the last line of defense)

dominykas: I hereby give such advice ;)

Then there's also https://blog.greenkeeper.io/introduction-to-semantic-release-33f73b117c8 and similar concepts.

I disagree that humans are better at this task than machines. I do however agree that this is not a one-sided decision, it has pros and cons. That said - I'm not sure that this document is a good place to weight them up, I think that's better left for CI/CD and publishing guidelines and I'm more than happy for those to include a balanced overview of both opinions.

So anyways, there is the school of thought of automated releases, there are maintainers using it, there are teams that can benefit and simplify their processes - they have no simple way of doing 2FA and that is the problem we're trying to solve here.

vweevers: Right, this document could focus on the _how_ of automating releases, not the _why_. It just needs a small tweak, because the current wording makes it sounds like automating the release process is considered a best practice by the package-maintenance team. Which - judging by @ljharb's comment - is not established, or even discussed yet.

dominykas: Sure, I'll see how to update the wording (happy to take suggestions).

Some prior discussions in #244.

The document title does only mention 2FA as a baseline practice.

dominykas: @vweevers does https://github.com/nodejs/package-maintenance/pull/282/commits/09035d7c5bdd5e82bec97f39596b452ebd6991ac sound better?

Also added the PR description.

vweevers: Perfect

docs/drafts/2fa.md
```diff
@@ -0,0 +1,167 @@
+# Baseline practice: using 2FA when publishing
+
+To avoid malicious use of stolen package registry tokens, it is advisable to enable two-factor authentication (2FA). When it is enabled every call to `npm publish` requires a unique one time password (OTP) to be entered.
+
+To make sure that the code is a _second_ factor it should be generated by a human on an independent device. The secret seed of the OTP generator should be protected in such a way, that it cannot be obtained together with the registry token.
+
+When releases are created by humans running `npm publish` (or via tooling such us [np](https://www.npmjs.com/package/np) or [publish-please](https://www.npmjs.com/package/publish-please)) on their local environments, using 2FA is not an issue.
```
wesleytodd: Maybe a better phrasing here would be "...on their local environments, using 2FA works as expected because the user is present to manually input the OTP."

Because you have not presented the issue yet, it is hard know here what "not an issue" means.  It kind of sounds like it is not necessary to have 2FA for this use case, which is *not* the intent.

docs/drafts/2fa.md
```diff
@@ -0,0 +1,167 @@
...
+
+Note: at the time of writing neither documents or gives advice on using two-factor authentication when publishing.
+
+## Potential solutions for 2FA in CI
+
+All of the listed solutions below have their pros and cons. While the order of desireability of these solutions has been discussed, the list below does not imply any order of preferences.
+
+### Release manager
+
+The release manager could be a place where packages are staged after `npm publish` without an OTP. A human would need to use the manager (CLI/GUI) and could release one or more pending packages, approving the publishing with an OTP. This would need be implemented by the registry (for public packages - npm).
```
wesleytodd: I think it is not required that this be implemented by the "registry" as we understand it today.  This could be a proxy service just as easily, operating as a gateway between publishers and the upstream registry.

wesleytodd (changes requested):
Sorry I only had time to read half right now (I started on a new team on Monday and have been very busy with the switch), but this is great!

docs/drafts/2fa.md
dominykas: I can extend the sentence here, sure, to include the "proxy" manager as an option, but it feels to me that such a proxy manager would work better and require less maintenance if it were CLI based tooling, using alternative registries/github releases as a backing store (i.e. last option). Maybe I should fold that last option into here?

If we want to publish to the public registry, then such a release manager would still need to have a way to provide 2FA for publishing things? In which case it is probably closer to using

docs/drafts/2fa.md
dominykas: Thanks, this makes sense. I removed the "not an issue" part altogether - does https://github.com/nodejs/package-maintenance/pull/282/commits/5db216160e89c81101db545c79296c58db00cd66 sound OK?

docs/drafts/2fa.md
wesleytodd: Yeah this is a hairy topic.  I think that as we move into a "multi-registry" world with GHPR there will be a bunch of new issues to tackle.  Handling all of them in the cli will be difficult for some issues.  I don't want to distract this post, which is great, with a discussion of alternate implementation ideas.  So I will resolve this conversation as it is, but I think we should remain open to different approaches and the language here is pretty restrictive.

docs/drafts/2fa.md
wesleytodd: Looks great!

boneskull:
This is a good overview of the problem, thanks

isaacs:
We just had a brief discussion on this topic in our npm team standup.

I intend to write up a brief RFC describing a staged publish support in the registry package document.  I'm going to avoid some of the various angles discussed here and try to keep it focused on what we can likely accomplish within a reasonable time window.

tl;dr of what I'm thinking:

- you can "stage" a publish by adding it to the "stagedPackages" object in the packument
- stagedPackages do not reserve the version number.  However, when a version is moved out of staging and into `versions`, the number is reserved, and any staged packages for that version are deleted.
- a new npm command to promote a staged publish to a real publish
- stagedPackages will be associated with a tarball artifact, so you can install them (for testing etc) by specifying the tarball URL.
- TBD:
    - permissions around who has rights to stage vs promote vs publish
    - possible npm option to include stagedVersions in install target list
    - Any UX (other than CLI and API) around release management, staging, etc.

I think this would give us a useful building block to start tackling some of the other stuff discussed here.

wesleytodd:
This sounds great @isaacs!  I think that what you have outlined here will really help, but the critical piece of integrating 2FA into the promote doesn't seem covered.  If you could do this on the registry website with a button click which required re-auth and 2FA that would be AWESOME!  So the sooner we can tackle that last TBD item the better :)

wesleytodd:
Hm, actually I wonder....maybe I was wrong that we need more than a cli.  I think having some user stories in the RFC would help here, because now that I am thinking about it you could have CI do the staged publish and then have the owner just run the promote from the cli with an OTP.  I guess that does solve that use case in a minimalist sense.

ljharb:
That's the workflow I envision; CI staging publishes, but the user explicitly promoting it (via CLI, and optionally in the future, the web UI)

docs/drafts/2fa.md
dominykas: Wrote https://github.com/nodejs/package-maintenance/pull/282/commits/1a609706193bdb9febdfa76134ff474c0130b511 before we got the response and forgot to push when I got back online.

darcyclarke:
As a follow up to @isaacs's initial comment, we've just posted an RFC outlining what that initial set of work/features might look like here: https://github.com/npm/rfcs/pull/92

ruyadorno:
@dominykas here's the link to the npm public events calendar that has the reference to each OpenRFC meeting: https://calendar.google.com/calendar/embed?src=npmjs.com_oonluqt8oftrt0vmgrfbg6q6go%40group.calendar.google.com (it's a bi-weekly happening meeting every other Wednesday)

dominykas:
Removed the agenda label until there's something more to discuss.

My plan is to wait for https://github.com/npm/rfcs/pull/92 and I will then rephrase this to list the _existing_ options as such, rather than the theoretical approaches on how this can be solved (might do that sooner, time permitting).

bnb:
May also want to consider adding something about npm access tokens.

https://docs.npmjs.com/creating-and-viewing-access-tokens

thescientist13:
These efforts are happening as well, maybe worth including or tracking in an issue to further refine this doc as progress is made
- https://github.blog/2022-02-01-top-100-npm-package-maintainers-require-2fa-additional-security/
- https://github.com/mmarchini-oss/npm-otp-publish

simoneb:
We have developed this which we're using to publish our npm packages with as much automation as having 2FA enabled in NPM allows. It's not a perfect solution but it works pretty well for us.

https://github.com/nearform/optic-release-automation-action

thescientist13 (approved):
Not sure we want to add a link to this to the [repo's docs TOC](https://github.com/nodejs/package-maintenance/tree/main/docs#table-of-contents) as well?

dominykas:
It seems staged publishing is no longer on the roadmap: https://github.com/npm/roadmap/issues/9#issuecomment-1065480566

I might actually see if I should start revisiting this doc sooner than I thought...

ljharb:
Yeah, it’s very disappointing, since it’s the only path to securely and reproducibly publishing anything.

OlawaleSC23 (approved):
@ @wesleytodd

Arefchab (approved):

simoneb:
Is this PR getting anywhere? It's been amusingly open for several years now

ljharb:
At this point, the only approach for actual 2FA in CI remains something like https://github.com/GoogleCloudPlatform/wombat-dressing-room or https://github.com/step-security/wait-for-secrets (the latter of which i use on eslint-plugin-react, the only package i publish from CI from so far).

With the "web" login, I suppose you could potentially remove the need for an external server, but I haven't investigated that yet.

dominykas:
I'm not sure I have enough capacity right now to rewrite any of the documentation here - these are mostly just design proposals.

In terms of implementation - given the age of this - I'm not sure any of the implementations are moving ahead or gaining significant traction to be called a common pattern... Maybe there's people somewhere else working on/discussing something similar?

Optic is as close as it gets (although it's grown up with a bunch more opinions), but the UX of all of these tools is just not there.

From my perspective, hitting the "Close" button on this PR is easy enough 🤷‍♂️ Maybe I should just do that?

wesleytodd:
> From my perspective, hitting the "Close" button on this PR is easy enough 🤷‍♂️ Maybe I should just do that?

I am not sure there is any benefit to doing that. At least with it being open that is a signal to new folks who might like to pick it up and get it updated. Maybe editing the original post with an update that it is "looking for help" and then you could just fun-ollow the thread?

Omw2fyb2 (approved):

docs/drafts/2fa.md
```diff
@@ -0,0 +1,167 @@
...
+- Can this be standardized enough to work with alternative registries?
+- This may required significant effort to build and maintain, incl. by npm Inc.
+
+Poor man's version:
+
+- Allow adding a 2FA approval on specific versions retroactively. This would allow tooling to be built on the consumer side to avoid non-2FA releases, while the maintainers keep the option of "singing off" the releases at their convenience - tooling can be built to help these tasks as well.
+
+### Extension in CI providers
+
+As 2FA in CI is partly a problem because there is no way to enter the OTP, it can be solved by simply adding a way to enter an OTP in CI.
```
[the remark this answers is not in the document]
pi0: Interesting idea but wouldn’t the comment be public?

(time based) OTPs can be used multiple times within valid time window. Bad actors can watch for comments and immediately reuse same OTP for another action.

Paidinful (approved):

PR #292, opened 2019-12-03
doc: fix link to LTS policy

mhdawson:


mhdawson:
@styfle that's true but not all links need to be in official Node.js repos.  If there are other definitions of support policies I'd be ok with linking to those.  I'm also open to a copy in the package-maintenance repo but I don't think we need to make copies.

mhdawson:
@nodejs/package-maintenance more approvals would be great so we can get the 7 day clock going.

Issue #299, opened 2019-12-13
[RFC] Workflow + Tooling for Package Maintainers 


thescientist13:
## Overview
For projects and packages operating within the NodeJS ecosystem (especially those with a high impact to the community), affording maintainers the tooling and workflows to confidently publish and distribute packages to their consumers has become an identifiable and growing desire for the community at large.

During the Collab Summit [session for the Package Maintenance WG](https://github.com/openjs-foundation/summit/issues/209) there was a great discussion around the topics of tooling, workflows, process, automation, etc and how these ideas can be explored and shared with the community at large, taking advantage of existing tooling already available.  (and maybe even making some new ones! 🎉 )

## Proposal (WIP)
The proposal here is to explore the opportunities discussed during the session and follow those threads where they may lead, in particular around finding opportunities around automation, tooling, and ease of use for maintainers.

Of great interest is cross pollinating tooling being developed and maintained by the NodeJS _Release_ Management WG, in particular tools like [**citgm**](https://github.com/nodejs/citgm), which s a tool the NodeJS teams uses to assess the impact of changes on the wider community at large, by building high impact projects _against_ new / nightly builds of NodeJS.

For example, a project like Express that is of high impact, may want to build against key projects and packages within its own ecosystem on a nightly basis to ensure there aren't breaking changes in a scenario outside of an expected semver major release.

## Thoughts / Questions / Considerations
- Given the high visibility and foundational support of NodeJS, it seems like there is a lot of opportunities to use and build off of what the NodeJS Release WG is already doing either by: 
increasing visibility of existing tools already used and automating more of the NodeJS process and sharing the results of those efforts with the community, ideally through CI.
- How do we measure and determine what's worth automating and also what could be directly into GitHub itself, like with GitHub actions, so that as many projects as possible can benefit with little lift / effort on their part, but with immediate value add to their project's maintenance.
- What expectations might from the community itself may we want to assert, e.g. if we build Express nightly, but all the ecosystem tools are mocking all the Express APIs, how effective will nightly **citgm** builds actually be?
- Are there opportunities for parallelization in our efforts here?  It would be great to think of a few different approaches and let teams / groups try them out and report back, and we can guide our efforts based on those results.

---

[...]

Issue #298, opened 2019-12-13
Follow up on npm project automation


wesleytodd:
At the collab summit rfc meeting @darcyclarke mentioned some work they are doing with automating things using probot.  Maybe we can collaborate on how to document your learnings?  I would love some of this for Express :)

jonchurch:
Closing as 🦗

Issue #280, opened 2019-11-09
[Baseline practices] Keeping dependencies and node up to date


dominykas:
We should have a doc, which describes the whys and the hows of staying up to date.

ghinks:
Something like the steps involved in setting up renovate bot on github for a repo and what are views are on the sort of schedule for updates?

dominykas:
Yes, although I think we should cover more than just renovate. There's also options for automerging, lock file maintenance, etc.

One thing I'd also be interested to see is keeping multiple release lines updated, esp. If we're going to make LTS for packages happen.

balupton:
FWIW

@bevry accomplishes the goal of this automatically for all of its packages, through the following pieces

- https://github.com/bevry/boundation enforces consistency against packages, including support for all the below
  - https://github.com/bevry/editions enables support for older node versions, as well as provides standardised data about the environments the package supports and what technology it uses
  - https://github.com/bevry/projectz enforces consistency of the meta files, such as the readme and package.json, and installation and license information
  - https://github.com/bevry/awesome-travis has several scripts such as
    - https://github.com/bevry/awesome-travis/blob/master/scripts/surge.bash which publishes the commit, branch, and tag to surge for browsing of docs and cdn
    - https://github.com/bevry/awesome-travis/blob/master/scripts/node-publish.bash publishes git tags to `latest` and master (as configured by `NPM_BRANCH_TAG`) to the `next` tag
    - https://github.com/bevry/awesome-travis/blob/master/scripts/node-install.bash ensures dev deps are installed with the desired node version, to enable tests to run even on legacy node versions
- https://github.com/balupton/dotfiles/blob/master/.scripts/users/balupton/commands/boundation-upgrade runs boundation on a project, and publishes a new release - if the engines changed, publishes a new major release, otherwise a minor
- https://github.com/balupton/dotfiles/blob/master/.scripts/users/balupton/commands/boundation-all runs boundation-upgrade on all bevry's packages

dominykas:
Adding to the meeting agenda:

I think we need an official repo (i.e. under `nodejs`, not `pkgjs` org) for Travis (and in the future - other CI setup?) includes. I know @ljharb was working on something - not sure how ready that is?

What I'd like to see is an include file with min version and the support policy - we can discuss details in the meeting/new issue (in the new repo?)

ljharb:
[...]

Issue #172, opened 2019-03-08
[best practices] managing LTS & release lines for userland modules


boneskull:
I'm starting a [discussion](/mochajs/mocha/issues/3819) in [Mocha](/mochajs/mocha) around how we can best support multiple "LTS" release lines.

I'm familiar with what Node.js itself does (though I haven't ever cut a Node.js release myself; it seems essentially like [trunk-based development](http://trunkbaseddevelopment.com).  But I don't have an understanding of all the gotchas around something like this.

It *does* seem that, without proper tooling, maintainers will be in for a lot of pain _in addition_ to the extra overhead that the strategy initially incurs.

I also don't know how other teams/projects that want to do long-lived release lines do it, and would love to hear from them!

How can we, as the package maintenance team, help projects that want to take this on?  What works and what doesn't?

boneskull:
(I wonder if @bethgriggs might have some insight?)

mcollina:
We are currently doing this in Fastify with v1 and v2. It’s not super straightforward, especially in handling backports in a semi-automated way.

dominykas:
I like how semantic-release (v16, beta, not latest) solves it.

It is rather opinionated, though, so might not always be feasible.

I'm curious to hear more about what seems to be the issue?

mhdawson:
@mcollina any details on what you are doing that you can share that might be input for this discussion ?

boneskull:
For automating backports, maybe something like a GH Action which runs a post-commit hook against a given branch (e.g., `master`) that will automatically cherry-pick the changeset(s) to release branches if it passes CI?  Otherwise maybe create an open PR someone can hack on?  Really wouldn't know if this sort of thing would be too burdensome or failure-prone until it was up & running...

ljharb:
In my experience, backports very rarely land cleanly, even if there's no git conflicts - it's always required a trivial bit of human tweaking.

mcollina:
In Fastify we are currently maintaining two release lines, v1 and v2. We have no automated tool for backporting. As such, we require a backport from the original author or the PR or from the developer landing the PR - or if there is a need.

It’s a lot free form, and a bit of automation will help.

wesleytodd:
> In my experience, backports very rarely land cleanly, even if there's no git conflicts - it's always required a trivial bit of human tweaking.

👍 Only trivial changes are able to be backported in an automated way.  I am not sure how much automation will be possible here, but would love to see a POC on the idea.

[...]

PR #300, opened 2019-12-14
tools.md: add bevry's automation tooling

### LINKED ISSUE #5: Tooling to assist maintenance
Travis CI recently informed me that I have over 800 packages using it.

I've done up the following tooling to assist in automating the maintenance of my packages as much as possible, they have been used for years, and probably others will find them useful:

- https://github.com/bevry/badges
- https://github.com/bevry/projectz
- https://github.com/bevry/boundation
- https://github.com/bevry/awesome-travis

balupton:
close #5

/cc #114 #172 #280 #298 #299


balupton:
In hindsight, I think `tools.md` should just be replaced by say a `awesome-maintenance` repo that follows in the awesome-* spirit.

ljharb:
why is this a standard we’re endorsing? I don’t think it’s appropriate for a node group to bless a particular community convention, even for an informational list.

wesleytodd:
TBQH, I think we should just delete this file until we have some clear purpose for it.  I suspect we will never want it as it would give too much weight even on the tools we end up building and supporting.

mhdawson:
I agree we should just delete the file.

mhdawson:
Going close this out since we are going to remove the file anyway.

Issue #301, opened 2019-12-18
Question: Is this part of a baseline practice? (link on description)


helio-frota:
Hi,

Not 100% clear for me but want to know if this is part of a baseline practice:

https://medium.com/@nodejs/choosing-the-node-js-versions-for-your-ci-tests-hint-use-lts-89b67f68d7ca

If not, no problem but is good to know anyway.

thanks

dominykas:
Yes, this recommendation is documented in https://github.com/nodejs/package-maintenance/blob/master/docs/drafts/testing-guidelines.md

helio-frota:
@dominykas thanks!

PR #302, opened 2019-12-23
Dependency management

### LINKED ISSUE #280: [Baseline practices] Keeping dependencies and node up to date

dominykas:
Closes #280.

I feel this is too long, but right now I'm not sure what can I leave out... There's probably some stuff which should just point to some external content.

Anything that's worth expanding on? Did I miss anything important?

### DIFF
--- a/docs/drafts/dependency-management-guidelines.md
+++ b/docs/drafts/dependency-management-guidelines.md
@@ -15,7 +15,7 @@ One of the basic maintenance tasks is ensuring that your package uses the latest
 There are some potential caveats to take into consideration when upgrading dependencies:
 
 - Dropping support for specific Node.js versions is normally considered a breaking change, which means that upgrading a dependency might imply that you also need to drop support for these versions, which in turn will force you to release a new major version of your package as well.
-- In some cases, upgrading a dependency might result in incompatibilities with other packages which rely on peer dependencies, i.e. you may be increasing the size of the `node_modules` folder and/or static bundle.
+- Upgrading a dependency could result in increased size of the end user's `node_modules` / compiled bundle, if other packages in the ecosystem do not upgrade around the same time.
 
 Node.js itself can also be considered a dependency of your package, therefore you should make sure to test in latest LTS and current Node.js versions in your Continuous Integration (CI) system - see [Choosing the Node.js version for testing](https://medium.com/@nodejs/choosing-the-node-js-versions-for-your-ci-tests-hint-use-lts-89b67f68d7ca).
 
--- a/docs/drafts/dependency-management-guidelines.md
+++ b/docs/drafts/dependency-management-guidelines.md
@@ -36,7 +36,7 @@ JavaScript ecosystem relies on [semantic versioning](https://semver.org/) and mo
 
 You may also use a tilde (`~`) character instead of the caret (`^`), to only accept bugfix (patch) releases, but this may mean that your dependency will become out of date sooner.
 
-Accepting _all_ releases (i.e. `*` instead of a version range) is risky and could break your package unexpectedly.
+Accepting _all_ releases (i.e. `*` or `latest` instead of a version range, or an unbounded version range using `>`/`>=`) is risky and almost certainly will break your package unexpectedly when a new major version of the dependency is released.
 
 Further reading:
 
--- a/docs/drafts/dependency-management-guidelines.md
+++ b/docs/drafts/dependency-management-guidelines.md
@@ -25,7 +25,7 @@ Node.js itself can also be considered a dependency of your package, therefore yo
 
 There is no single recipe for choosing the right dependency for your package, but there are some indicators to pay attention to and questions to ask yourself:
 
-- Do you need a dependency at all? Each dependency has a download cost, and if it is build by someone else - it may have security and reliability implications.
+- Should you build it yourself or use a package from the ecosystem? Adding a dependency is a trade-off. Each dependency has a download cost, and if it is built by someone else - it may have security and reliability implications. However building your own implementation may suffer from even worse bugs (incl. security) and might not have enough eyes to catch them. 
 - Does the package have tests and do they run automatically in a CI system, e.g. Travis or GitHub Actions.
 - Is the package actively maintained? Lack of activity in the source control does not mean it is unmaintained, as it may simply be done, but it is something to pay attention to.
 - Make sure the package is not already marked as deprecated on npm and in the README - authors may give you suggestions for alternatives.
--- a/docs/drafts/dependency-management-guidelines.md
+++ b/docs/drafts/dependency-management-guidelines.md
@@ -6,23 +6,22 @@ This document outlines the best practices on keeping package dependencies (incl.
 
 ## Why
 
-One of the basic maintenance tasks is ensuring that the package uses the latest stable versions of its dependencies. Here are some reasons why it's a good practice to do so:
+One of the basic maintenance tasks is ensuring that your package uses the latest stable versions of its dependencies. There are multiple benefits of doing this:
 
-- Having latest bugfixes / features
-- Software release lifecycle
-- Security
+- The primary reason to keep your dependencies up to date is to receive all the latest bugfixes and improvements.
+- Old versions of your dependencies might no longer be receiving security updates, which means that if a new serious vulnerability is disclosed, you may be forced to do extra unplanned work if you are multiple versions behind.
+- Upgrading regularly in small incremental steps is potentially less disruptive than doing a big bang update to adapt to many breaking changes released over multiple versions of a dependency.
 
-Potential caveats to take into consideration:
+There are some potential caveats to take into consideration when upgrading dependencies:
 
-- Node.js version support
-- Effects on the peer dependencies, incl. `node_modules` bloat
+- Dropping support for specific Node.js versions is normally considered a breaking change, which means that upgrading a dependency might imply that you also need to drop support for these versions, which in turn will force you to release a new major version of your package as well.
+- In some cases, upgrading a dependency might result in incompatibilities with other packages which rely on peer dependencies, i.e. you may be increasing the size of the `node_modules` folder and/or static bundle.
 
-It is also important to make sure to test with latest LTS and current Node.js versions in CI/CD.
-
-- [Choosing the Node.js version for testing](https://medium.com/@nodejs/choosing-the-node-js-versions-for-your-ci-tests-hint-use-lts-89b67f68d7ca)
+Node.js itself can also be considered a dependency of your package, therefore you should make sure to test in latest LTS and current Node.js versions in your Continuous Integration system - see [Choosing the Node.js version for testing](https://medium.com/@nodejs/choosing-the-node-js-versions-for-your-ci-tests-hint-use-lts-89b67f68d7ca).
 
 ## How
 
+- Choosing dependencies
 - Defining version ranges in `package.json`
 - Using `npm outdated`
 - Using `npm update`
--- a/docs/drafts/dependency-management-guidelines.md
+++ b/docs/drafts/dependency-management-guidelines.md
@@ -2,7 +2,7 @@
 
 ## What
 
-This document outlines the best practices on keeping package dependencies (incl. Node itself) up to date.
+This document outlines the best practices on keeping package dependencies (including Node itself) up to date.
 
 ## Why
 
--- a/docs/drafts/dependency-management-guidelines.md
+++ b/docs/drafts/dependency-management-guidelines.md
@@ -29,6 +29,8 @@ There is no single recipe for choosing the right dependency for your package, bu
 - Does the package have tests and do they run automatically in a CI system, e.g. Travis or GitHub Actions.
 - Is the package actively maintained? Lack of activity in the source control does not mean it is unmaintained, as it may simply be done, but it is something to pay attention to.
 - Make sure the package is not already marked as deprecated on npm and in the README - authors may give you suggestions for alternatives.
+- Does the project offer support? Do the releases have a changelog to ease the update?
+  
 
 ### Defining the dependency version ranges
 
--- a/docs/drafts/dependency-management-guidelines.md
+++ b/docs/drafts/dependency-management-guidelines.md
@@ -40,6 +40,8 @@ You may also use a tilde (`~`) character instead of the caret (`^`), to only acc
 
 Accepting _all_ releases (i.e. `*` or `latest` instead of a version range, or an unbounded version range using `>`/`>=`) is risky and almost certainly will break your package unexpectedly when a new major version of the dependency is released.
 
+As with adding dependencies in general, updating them without any review carries security risks. In sensitive contexts you might want to consider gatekeeping every update by pinning the dependencies to their exact version.
+
 Further reading:
 
 - npm documentation: [About semantic version](https://docs.npmjs.com/about-semantic-versioning)
--- a/docs/drafts/dependency-management-guidelines.md
+++ b/docs/drafts/dependency-management-guidelines.md
@@ -83,3 +83,6 @@ All of the below tools provide a GitHub App to create PRs with updates. They als
 - [Renovate](https://renovate.whitesourcesoftware.com/)
     - Supports other languages, besides Javascript
     - Supports updating `.travis.yml`, Circle CI yml and `.nvmrc`, i.e. it can update the Node.js version you use for running tests
+- [Snyk](https://snyk.io/)
+    - Supports other languages, besides JavaScript
+    - Supports ignoring specific deps, and limiting the amount of open simultaneous PRs to reduce noise level
--- a/docs/drafts/dependency-management-guidelines.md
+++ b/docs/drafts/dependency-management-guidelines.md
@@ -81,7 +81,7 @@ All of the below tools provide a GitHub App to create PRs with updates. They als
 - [Greenkeeper](https://greenkeeper.io/)
     - Creates branches to test even the in-range updates and reports status if they make tests fail
 - [Renovate](https://renovate.whitesourcesoftware.com/)
-    - Supports other languages, besides Javascript
+    - Supports other languages, besides JavaScript
     - Supports updating `.travis.yml`, Circle CI yml and `.nvmrc`, i.e. it can update the Node.js version you use for running tests
 - [Snyk](https://snyk.io/)
     - Supports other languages, besides JavaScript
--- a/docs/drafts/dependency-management-guidelines.md
+++ b/docs/drafts/dependency-management-guidelines.md
@@ -59,6 +59,7 @@ Using the lock files has downsides:
 
 - `package-lock.json` (and `yarn.lock`) is only used in development, so you may be testing your package with a different set of dependency versions than your users will be using it, which may occasionally lead to unexpected results.
 - no dependency updates will be received automatically with `npm install`.
+- using a shrinkwrap in a package intended to be used as a library may prevent deduplication of dependencies, increasing the `node_modules` / bundle size.
 
 You can disable the lock files in your global or local `.npmrc` file by adding a `package-lock=false` line. This does however mean that npm will also ignore the shrinkwraps included in your dependencies.
 
--- a/docs/drafts/dependency-management-guidelines.md
+++ b/docs/drafts/dependency-management-guidelines.md
@@ -45,7 +45,22 @@ Further reading:
 
 ### Using lock files
 
-(todo)
+The lock files are a snapshot of your full dependency tree, incl. all sub-dependencies, their precise versions and also integrity hashes. Having a lock file allows npm to recreate the dependency tree exactly as it was. npm supports two lock files: `package-lock.json` and `npm-shrinkwrap.json`. Yarn uses its own lock file format.
+
+The `package-lock.json` will get created automatically by default and it will not be published into the registry, i.e. it is only intended to be used for development purposes. npm recommends you commit this file into your source control.
+
+You can chose to use the `npm shrinkwrap` command to instead create an `npm-srhinkwrap.json` file. It is the same in structure as the `package-lock.json`, but it will also be published together with your package in the registry, which means that not only people who develop the package, but also the users will receive the exact set of dependencies.
+
+Using the lock files has downsides:
+
+- `package-lock.json` is only used in development, so you may be testing your package with a different set of dependency versions than your users will be using it, which may occasionally lead to unexpected results.
+- no dependency updates will be received automatically with `npm install`.
+
+You can disable the lock files in your global or local `.npmrc` file by adding a `package-lock=false` line. This does however mean that npm will also ignore the shrinkwraps included in your dependencies.
+
+Further reading:
+
+- npm documentation: [npm-package-locks](https://docs.npmjs.com/files/package-locks)
 
 ### Updating the dependencies manually
 
--- a/docs/drafts/dependency-management-guidelines.md
+++ b/docs/drafts/dependency-management-guidelines.md
@@ -17,16 +17,48 @@ There are some potential caveats to take into consideration when upgrading depen
 - Dropping support for specific Node.js versions is normally considered a breaking change, which means that upgrading a dependency might imply that you also need to drop support for these versions, which in turn will force you to release a new major version of your package as well.
 - In some cases, upgrading a dependency might result in incompatibilities with other packages which rely on peer dependencies, i.e. you may be increasing the size of the `node_modules` folder and/or static bundle.
 
-Node.js itself can also be considered a dependency of your package, therefore you should make sure to test in latest LTS and current Node.js versions in your Continuous Integration system - see [Choosing the Node.js version for testing](https://medium.com/@nodejs/choosing-the-node-js-versions-for-your-ci-tests-hint-use-lts-89b67f68d7ca).
+Node.js itself can also be considered a dependency of your package, therefore you should make sure to test in latest LTS and current Node.js versions in your Continuous Integration (CI) system - see [Choosing the Node.js version for testing](https://medium.com/@nodejs/choosing-the-node-js-versions-for-your-ci-tests-hint-use-lts-89b67f68d7ca).
 
 ## How
 
-- Choosing dependencies
-- Defining version ranges in `package.json`
-- Using `npm outdated`
-- Using `npm update`
-- Using lockfiles
-- Tools to automatically manage updates
-    - Greenkeeper
-    - Renovate
-    - Dependabot
+### Choosing dependencies
+
+There is no single recipe for choosing the right dependency for your package, but there are some indicators to pay attention to and questions to ask yourself:
+
+- Do you need a dependency at all? Each dependency has a download cost, and if it is build by someone else - it may have security and reliability implications.
+- Does the package have tests and do they run automatically in a CI system, e.g. Travis or GitHub Actions.
+- Is the package actively maintained? Lack of activity in the source control does not mean it is unmaintained, as it may simply be done, but it is something to pay attention to.
+- Make sure the package is not already marked as deprecated on npm and in the README - authors may give you suggestions for alternatives.
+
+### Defining the dependency version ranges
+
+JavaScript ecosystem relies on [semantic versioning](https://semver.org/) and most packages respect its rules. Therefore, it is usually recommended that your `package.json` accepts new minor/patch releases of your dependencies automatically. This is the default behavior of npm - when you first `npm install [new dependency]`, it will by default prefix the version with a `^` to indicate that all new releases under the same major version are acceptable.
+
+You may also use a tilde (`~`) character instead of the caret (`^`), to only accept bugfix (patch) releases, but this may mean that your dependency will become out of date sooner.
+
+Accepting _all_ releases (i.e. `*` instead of a version range) is risky and could break your package unexpectedly.
+
+Further reading:
+
+- npm documentation: [About semantic version](https://docs.npmjs.com/about-semantic-versioning)
+- npm documentation: [Updating packages downloaded from the registry](https://docs.npmjs.com/updating-packages-downloaded-from-the-registry)
+
+### Using lock files
+
+(todo)
+
+### Updating the dependencies manually
+
+You can list outdated dependencies using the `npm outdated` command. If there are any in-range updates - you can use `npm update` command to install them.
+
+You will need to modify your `package.json` version specifier to receive out-of-range updates.
+
+There are several tools that will help you manage this from your local command line, such as [`npm-check`](https://www.npmjs.com/package/npm-check) and [`npm-check-updates`](https://www.npmjs.com/package/npm-check-updates), however you can also set up external tools to automatically create Pull Requests with updates. 
+
+### Automatically keeping dependencies up to date
+
+(todo)
+
+- Greenkeeper
+- Renovate
+- Dependabot
--- a/docs/drafts/dependency-management-guidelines.md
+++ b/docs/drafts/dependency-management-guidelines.md
@@ -49,7 +49,7 @@ The lock files are a snapshot of your full dependency tree, incl. all sub-depend
 
 The `package-lock.json` will get created automatically by default and it will not be published into the registry, i.e. it is only intended to be used for development purposes. npm recommends you commit this file into your source control.
 
-You can chose to use the `npm shrinkwrap` command to instead create an `npm-srhinkwrap.json` file. It is the same in structure as the `package-lock.json`, but it will also be published together with your package in the registry, which means that not only people who develop the package, but also the users will receive the exact set of dependencies.
+You can chose to use the `npm shrinkwrap` command to instead create an `npm-shrinkwrap.json` file. It is the same in structure as the `package-lock.json`, but it will also be published together with your package in the registry, which means that not only people who develop the package, but also the users will receive the exact set of dependencies.
 
 Using the lock files has downsides:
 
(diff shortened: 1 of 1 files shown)


helio-frota:
Thanks for this one.
Also good to see the blog post ( that I've questioned here https://github.com/nodejs/package-maintenance/issues/301. ) as part of the document.

Just to let you know, we updated repositories in two github organizations to follow that practice related to CI.

https://github.com/nodeshift
https://github.com/nodeshift-starters

And looks like we don't need to worry about update Node.js in CI (.travis.yml) until April 2021 ( for our use case ) :smile:

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+
+All of the below tools provide a GitHub App to create PRs with updates. They also support maintaining lock files to update sub-dependencies. Each tool also has built-in or community supported ways of automatically merging the PRs as long as tests pass, although this does carry some risk.
+
+- [Dependabot](https://dependabot.com/)
+    - Supports other languages, besides JavaScript 
+- [Greenkeeper](https://greenkeeper.io/)
+    - Creates branches to test even the in-range updates and reports status if they make tests fail
+- [Renovate](https://renovate.whitesourcesoftware.com/)
+    - Supports other languages, besides Javascript
+    - Supports updating `.travis.yml`, Circle CI yml and `.nvmrc`, i.e. it can update the Node.js version you use for running tests
```
lirantal:
```suggestion
    - Supports updating `.travis.yml`, Circle CI yml and `.nvmrc`, i.e. it can update the Node.js version you use for running tests
- [Snyk](https://snyk.io/)
    - Supports other languages, besides JavaScript
    - Supports ignoring specific deps, and limiting the amount of open simultaneous PRs to reduce noise level
```

lirantal: added Snyk too to the list which has dep ugprades

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+If your package has [automated tests](testing-guidelines.md) and [CI/CD](ci-cd-guidelines.md), you can use one of several services, free for Open Source Software, to automatically keep your dependencies up to date.
+
+All of the below tools provide a GitHub App to create PRs with updates. They also support maintaining lock files to update sub-dependencies. Each tool also has built-in or community supported ways of automatically merging the PRs as long as tests pass, although this does carry some risk.
+
+- [Dependabot](https://dependabot.com/)
+    - Supports other languages, besides JavaScript 
+- [Greenkeeper](https://greenkeeper.io/)
+    - Creates branches to test even the in-range updates and reports status if they make tests fail
+- [Renovate](https://renovate.whitesourcesoftware.com/)
+    - Supports other languages, besides Javascript
```
lirantal:
```suggestion
    - Supports other languages, besides JavaScript
```

lirantal: the one that got away ;-)

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,85 @@
+# Dependency management
+
+## What
+
+This document outlines the best practices on keeping package dependencies (incl. Node itself) up to date.
```
ljharb: we're not paying by the letter, let's avoid abbreviations, for clarity :-)
```suggestion
This document outlines the best practices on keeping package dependencies (including Node itself) up to date.
```

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+
+- The primary reason to keep your dependencies up to date is to receive all the latest bugfixes and improvements.
+- Old versions of your dependencies might no longer be receiving security updates, which means that if a new serious vulnerability is disclosed, you may be forced to do extra unplanned work if you are multiple versions behind.
+- Upgrading regularly in small incremental steps is potentially less disruptive than doing a big bang update to adapt to many breaking changes released over multiple versions of a dependency.
+
+There are some potential caveats to take into consideration when upgrading dependencies:
+
+- Dropping support for specific Node.js versions is normally considered a breaking change, which means that upgrading a dependency might imply that you also need to drop support for these versions, which in turn will force you to release a new major version of your package as well.
+- In some cases, upgrading a dependency might result in incompatibilities with other packages which rely on peer dependencies, i.e. you may be increasing the size of the `node_modules` folder and/or static bundle.
```
ljharb: can you elaborate on this one?

If a dep is something that should be a peer dep, then it should be one everywhere, and any peer dep change that causes a previously valid version to be invalid is a breaking change. If it's not a peer dep, there should be no incompatibilities.

I'm not aware of anything that should be a peer dep sometimes, and a regular dep other times, except in top-level code.

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+
+Node.js itself can also be considered a dependency of your package, therefore you should make sure to test in latest LTS and current Node.js versions in your Continuous Integration (CI) system - see [Choosing the Node.js version for testing](https://medium.com/@nodejs/choosing-the-node-js-versions-for-your-ci-tests-hint-use-lts-89b67f68d7ca).
+
+## How
+
+### Choosing dependencies
+
+There is no single recipe for choosing the right dependency for your package, but there are some indicators to pay attention to and questions to ask yourself:
+
+- Do you need a dependency at all? Each dependency has a download cost, and if it is build by someone else - it may have security and reliability implications.
```
ljharb: I very much object to discouraging people from adding a dependency - one that is likely to be better tested and more widely used than your own alternative - based on *download cost*. Separately, first-party code has imo *more* risk in terms of security and reliability than most third party code.

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+- Is the package actively maintained? Lack of activity in the source control does not mean it is unmaintained, as it may simply be done, but it is something to pay attention to.
+- Make sure the package is not already marked as deprecated on npm and in the README - authors may give you suggestions for alternatives.
+
+### Defining the dependency version ranges
+
+JavaScript ecosystem relies on [semantic versioning](https://semver.org/) and most packages respect its rules. Therefore, it is usually recommended that your `package.json` accepts new minor/patch releases of your dependencies automatically. This is the default behavior of npm - when you first `npm install [new dependency]`, it will by default prefix the version with a `^` to indicate that all new releases under the same major version are acceptable.
+
+You may also use a tilde (`~`) character instead of the caret (`^`), to only accept bugfix (patch) releases, but this may mean that your dependency will become out of date sooner.
+
+Accepting _all_ releases (i.e. `*` instead of a version range) is risky and could break your package unexpectedly.
```
ljharb:
```suggestion
Accepting _all_ releases (i.e. `*` or `latest` instead of a version range, or an unbounded version range using `>`/`>=`) is risky and almost certainly will break your package unexpectedly when a new major version of the dependency is released.
```
I think we could use even stronger language here; nobody should ever do this.

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+
+The lock files are a snapshot of your full dependency tree, incl. all sub-dependencies, their precise versions and also integrity hashes. Having a lock file allows npm to recreate the dependency tree exactly as it was. npm supports two lock files: `package-lock.json` and `npm-shrinkwrap.json`. Yarn uses its own lock file format.
+
+The `package-lock.json` will get created automatically by default and it will not be published into the registry, i.e. it is only intended to be used for development purposes. npm recommends you commit this file into your source control.
+
+You can chose to use the `npm shrinkwrap` command to instead create an `npm-srhinkwrap.json` file. It is the same in structure as the `package-lock.json`, but it will also be published together with your package in the registry, which means that not only people who develop the package, but also the users will receive the exact set of dependencies.
+
+Using the lock files has downsides:
+
+- `package-lock.json` is only used in development, so you may be testing your package with a different set of dependency versions than your users will be using it, which may occasionally lead to unexpected results.
```
ljharb:
```suggestion
- `package-lock.json` (and `yarn.lock`) is only used in development, so you may be testing your package with a different set of dependency versions than your users will be using it, which may occasionally lead to unexpected results.
```

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+- no dependency updates will be received automatically with `npm install`.
+
+You can disable the lock files in your global or local `.npmrc` file by adding a `package-lock=false` line. This does however mean that npm will also ignore the shrinkwraps included in your dependencies.
```
ljharb:
```suggestion
You can disable npm's default lockfile in your global or local `.npmrc` file by adding a `package-lock=false` line.
```
Removed the second sentence as afaik it is quite incorrect; if *anything* causes npm (or yarn) to ignore a published shrinkwrap file, npm is broken, and we should file an issue instead.

docs/drafts/dependency-management-guidelines.md
dominykas: I understand the sentiment and I'm also a fan of small modules and doing one thing well, however this is not a one sided argument. But you're right - it's lazy of me to word this as a "download cost" problem alone. The security implications are a trade off and depend on the use case, e.g. there's no point to drag in the whole of `chalk` if you just want to make one line of text red and I would strongly disagree that looking for a module which only does the red colouring is worth it either.

I would also strongly disagree on the first vs third party code - once again - it depends case-by-case, and there's authors I trust more than others, and I'd never write my own crypto, but at the same time - we _have_ seen incidents, fairly recently, and the risk is there, so a piece of code that one has written themselves, for a specific use case, where they understand it 100% is easier to manage than to try to track every possible change in a dependency, which is likely not even related.

I'm sure we could debate this for ages.

I'll rephrase this to give a better overview of both perspectives.

docs/drafts/dependency-management-guidelines.md
dominykas: This has worked like that since `package-lock.json` was introduced though? 

[Docs for `package-lock`](https://docs.npmjs.com/misc/config#package-lock) say:

> This option is an alias for --shrinkwrap.

And setting it to `false` disables both - creation and usage of all the lock files. 

I definitely think this behavior is not something I expect or find intuitive and it gets worse when you have `package-lock` and `shrinkwrap` options set to different values, so maybe yes - an issue on npm tracker is in order (if nobody has one created yet), but I'd say it is worth documenting this (with a link to the issue) until it is fixed?

docs/drafts/dependency-management-guidelines.md
dominykas: Sure, I'll revisit and rephrase if necessary. I had it in my notes as peer dep, but I can only dig up an example with regular deps bloating the bundles right now, so I need to look further.

dominykas: On a somewhat related note, I wonder if there's a place in this doc to explain the gotchas of peer deps, or whether it would work better a separate doc 🤔

docs/drafts/dependency-management-guidelines.md
ljharb: It disables creation but as far as I'm aware does not disable usage of nested npm-shrinkwrap files; a repro case would be interesting.

Either way, it's user-hostile to publish a shrinkwrap file (since it prevents deduping) so i doubt that it's a common enough case to warrant mentioning.

docs/drafts/dependency-management-guidelines.md
ljharb: > there's no point to drag in the whole of chalk if you just want to make one line of text red

I find plenty of point in doing it; that chalk does more than you need doesn't make it easy to correctly do the subset. For example, virtually every single complainer about "leftpad" who offered a one-liner had subtle bugs (i was moving String.prototype.padStart through TC39 at the time).

> I'll rephrase this to give a better overview of both perspectives.

Thanks, I think a balanced discussion of tradeoffs is reasonable

docs/drafts/dependency-management-guidelines.md
ljharb: Peer deps are sufficiently complex that hardly anybody fully understands them, including the maintainers of JS package managers; a separate doc seems best.

docs/drafts/dependency-management-guidelines.md
dominykas: [Docs for `shrinkwrap`](https://docs.npmjs.com/misc/config#shrinkwrap) say:

> If set to false, then ignore npm-shrinkwrap.json files when installing.

I'd have to double check, but I think it ignores the published shrinkwraps as well as the ones present in the folder next to package.json.

> it's user-hostile to publish a shrinkwrap file (since it prevents deduping)

I don't think many people do that, and even hapi gave up the practice, but if the shrinkwraps are there - they will be ignored and it is not necessarily something people will expect?

ljharb: Perhaps, but i doubt the installer of a package has any expectation that there's a shrinkwrap file there - it seems worth mentioning to package authors who publish a shrinkwrap file, but I'd prefer to use that space to tell them not to publish one in the first place :-)

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,88 @@
...
+
+## How
+
+### Choosing dependencies
+
+There is no single recipe for choosing the right dependency for your package, but there are some indicators to pay attention to and questions to ask yourself:
+
+- Do you need a dependency at all? Each dependency has a download cost, and if it is build by someone else - it may have security and reliability implications.
+- Does the package have tests and do they run automatically in a CI system, e.g. Travis or GitHub Actions.
+- Is the package actively maintained? Lack of activity in the source control does not mean it is unmaintained, as it may simply be done, but it is something to pay attention to.
```
Eomm: Could we add a memo to check if the maintainers publish also a changelog? 

It happens to me several times to check the commit messages between releases

dominykas: Can you suggest the wording?

Eomm: I would say something like:

Does the project offer support? Do the releases have a changelog to easy the update?

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,88 @@
...
+- `package-lock.json` (and `yarn.lock`) is only used in development, so you may be testing your package with a different set of dependency versions than your users will be using it, which may occasionally lead to unexpected results.
+- no dependency updates will be received automatically with `npm install`.
+
+You can disable the lock files in your global or local `.npmrc` file by adding a `package-lock=false` line. This does however mean that npm will also ignore the shrinkwraps included in your dependencies.
+
+Further reading:
+
+- npm documentation: [npm-package-locks](https://docs.npmjs.com/files/package-locks)
+
+### Updating the dependencies manually
```
dominykas: Note to self: add more about security implications.

docs/drafts/dependency-management-guidelines.md
dominykas: Checked a specific I had in mind - it wasn't about peer deps, just a general inability to deduplicate when several related packages where using different majors of a dep. Reworded: https://github.com/nodejs/package-maintenance/pull/302/commits/5f0c56c25d773e927537dbb9c9ec341c4f6cbbb0

docs/drafts/dependency-management-guidelines.md
dominykas: https://github.com/nodejs/package-maintenance/pull/302/commits/fd3c19c79df75a0bd124590b184ae3b4722bc878

docs/drafts/dependency-management-guidelines.md
dominykas: OK, so I did a quick test with latest npm. When `package-lock=false` and a dependency has a shrinkwrap inside the tarball, the shrinkwrap is ignored. So I think the original statement is correct - `package-lock=false` disables _all_ lockfiles - they will not be respected, nor will they be created, so the suggestion is not entirely correct and I think it's worth keeping the note that shrinkwraps will be ignored.

I've added a bullet point on deduplication: https://github.com/nodejs/package-maintenance/pull/302/commits/c126a7d3a06baad34e5e231457e4136388772be2

docs/drafts/dependency-management-guidelines.md
mhdawson: Agreed that the discussion should include more than "download cost".  Things like the licence requirements, and level of ongoing maintenance/support is more important to me.  I do agree though that some thought to the size of what you are pulling in does makes sense as well.

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Choosing dependencies
+
+There is no single recipe for choosing the right dependency for your package, but there are some indicators to pay attention to and questions to ask yourself:
+
+- Do you need a dependency at all? Each dependency has a download cost, and if it is build by someone else - it may have security and reliability implications.
+- Does the package have tests and do they run automatically in a CI system, e.g. Travis or GitHub Actions.
+- Is the package actively maintained? Lack of activity in the source control does not mean it is unmaintained, as it may simply be done, but it is something to pay attention to.
+- Make sure the package is not already marked as deprecated on npm and in the README - authors may give you suggestions for alternatives.
+- Does the project offer support? Do the releases have a changelog to ease the update?
```
mhdawson: Just a note for later, this might be a place to reference the support info (once we start to see adoption) as one way to figure this out.

docs/drafts/dependency-management-guidelines.md
dominykas: Does https://github.com/nodejs/package-maintenance/pull/302/commits/acdd9c561f51cc1f36af6f56b55d22193693f18a sound ok?

ljharb: Somewhat - it doesn't have implications *because* it was built by somebody else, it's that if you didn't build it, you have to audit and understand it in order to evaluate those possible implications.

In all cases, you have to understand the risks and understand what the code is doing; it's just that if you do it yourself, you're not sharing that risk with anyone else. In other words, I think that there is inherently *more* risk in reinventing wheels, and I'd like to see prose that speaks to that.

dominykas: Can you suggest improvements to the existing text?

ljharb: I think anything that implies that avoiding a dependency is something to strive for is inherently problematic.

How about replacing this item with some combination of:
Is this dependency the right tool for the job? Does it appear well-maintained, do you trust the authors/maintainers? Does it have thorough automated test coverage? If you had to fix a bug in it, could you? Is the problem it solves encapsulated well, so that it would be easy to replace this dependency with another if needed?

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,93 @@
...
+### Automatically keeping dependencies up to date
+
+If your package has [automated tests](testing-guidelines.md) and [CI/CD](ci-cd-guidelines.md), you can use one of several services, free for Open Source Software, to automatically keep your dependencies up to date.
+
+All of the below tools provide a GitHub App to create PRs with updates. They also support maintaining lock files to update sub-dependencies. Each tool also has built-in or community supported ways of automatically merging the PRs as long as tests pass, although this does carry some risk.
+
+- [Dependabot](https://dependabot.com/)
+    - Supports other languages, besides JavaScript 
+- [Greenkeeper](https://greenkeeper.io/)
+    - Creates branches to test even the in-range updates and reports status if they make tests fail
```
richardlau, line 87: FWIW Greenkeeper is being sunset on 3rd June and are no longer allowing new signups.

dominykas: Thanks, I'll update this.

ljharb: it's being replaced with a Snyk service, however, so we'll probably want to mention that instead.

nschonni: Snyk is already mentioned below below

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,93 @@
...
+
+There are several tools that will help you manage this from your local command line, such as [`npm-check`](https://www.npmjs.com/package/npm-check) and [`npm-check-updates`](https://www.npmjs.com/package/npm-check-updates), however you can also set up external tools to automatically create Pull Requests with updates. 
+
+### Automatically keeping dependencies up to date
+
+If your package has [automated tests](testing-guidelines.md) and [CI/CD](ci-cd-guidelines.md), you can use one of several services, free for Open Source Software, to automatically keep your dependencies up to date.
+
+All of the below tools provide a GitHub App to create PRs with updates. They also support maintaining lock files to update sub-dependencies. Each tool also has built-in or community supported ways of automatically merging the PRs as long as tests pass, although this does carry some risk.
+
+- [Dependabot](https://dependabot.com/)
```
sam-github, line 84: This was acquired by github, as greenkeeper was by snyk, some consolidation is occuring.

docs/drafts/dependency-management-guidelines.md
dominykas: I'll remove the Greenkeeper section, but I'm quite concerned about this bit:

> It does not require or directly engage with a CI tool.

Which basically means, that there is no more tool, which can run tests with new deps, and raise alerts that your CI won't pass even with an in-range update?

I'm sure someone will come along and create such a service, but for now - even though my personal preference is not GK - that feels like a major loss of functionality?

ljharb: I was concerned about that tool, as a dep updating tool that doesn't run CI is basically useless.

The impression folks at Snyk gave me was that this was an error in greenkeeper's post about the topic, and that it would, just like greenkeeper, invoke your CI - but it'd be good to get explicit confirmation.

dominykas: @lirantal - can you suggest a new wording for post-greenkeeper snyk?

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,93 @@
...
+- npm documentation: [About semantic version](https://docs.npmjs.com/about-semantic-versioning)
+- npm documentation: [Updating packages downloaded from the registry](https://docs.npmjs.com/updating-packages-downloaded-from-the-registry)
+
+### Using lock files
+
+The lock files are a snapshot of your full dependency tree, incl. all sub-dependencies, their precise versions and also integrity hashes. Having a lock file allows npm to recreate the dependency tree exactly as it was. npm supports two lock files: `package-lock.json` and `npm-shrinkwrap.json`. Yarn uses its own lock file format.
+
+The `package-lock.json` will get created automatically by default and it will not be published into the registry, i.e. it is only intended to be used for development purposes. npm recommends you commit this file into your source control.
+
+You can chose to use the `npm shrinkwrap` command to instead create an `npm-shrinkwrap.json` file. It is the same in structure as the `package-lock.json`, but it will also be published together with your package in the registry, which means that not only people who develop the package, but also the users will receive the exact set of dependencies.
```
sam-github, line 56: npm-shrinkwrap.json is created by default now

sam-github, line 56: "also the users will receive the exact set of dependencies" - not usually, since most packages in npmjs.com are installed as dependencies, and shrinkwraps are ignored for package dependencies. The only time its worth doing this is when publishing standalone apps/clis, not packages that are intended to be used as deps.

ljharb, line 56:
```suggestion
You can choose to use the `npm shrinkwrap` command to instead create an `npm-shrinkwrap.json` file. It is the same in structure as the `package-lock.json`, but it will also be published together with your package in the registry, which means that not only people who develop the package, but also the users will receive the exact set of dependencies, and will be unable to dedupe transitive dependencies with the rest of their dependency graph.
```

dominykas: A note on this is already included under the downsides?

ljharb: I don't think we can over-warn against including a shrinkwrap file in packages.

thescientist13 (commented):
In the spirit of #373 does anyone have any objections to having this just be in _docs/_, as opposed to be a draft in _docs/drafts/_?

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,93 @@
...
+- Does the package have tests and do they run automatically in a CI system, e.g. Travis or GitHub Actions.
+- Is the package actively maintained? Lack of activity in the source control does not mean it is unmaintained, as it may simply be done, but it is something to pay attention to.
+- Make sure the package is not already marked as deprecated on npm and in the README - authors may give you suggestions for alternatives.
+- Does the project offer support? Do the releases have a changelog to ease the update?
+  
+
+### Defining the dependency version ranges
+
+JavaScript ecosystem relies on [semantic versioning](https://semver.org/) and most packages respect its rules. Therefore, it is usually recommended that your `package.json` accepts new minor/patch releases of your dependencies automatically. This is the default behavior of npm - when you first `npm install [new dependency]`, it will by default prefix the version with a `^` to indicate that all new releases under the same major version are acceptable.
```
ljharb, line 37:
```suggestion
The JavaScript ecosystem relies on [semantic versioning](https://semver.org/) and most packages respect its rules. Therefore, it is usually recommended that your `package.json` accepts new minor/patch releases of your dependencies automatically. This is the default behavior of npm - when you first `npm install [new dependency]`, it will by default prefix the version with a `^` to indicate that all new releases under the same major version are acceptable.
```

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,93 @@
...
+As with adding dependencies in general, updating them without any review carries security risks. In sensitive contexts you might want to consider gatekeeping every update by pinning the dependencies to their exact version.
+
+Further reading:
+
+- npm documentation: [About semantic version](https://docs.npmjs.com/about-semantic-versioning)
+- npm documentation: [Updating packages downloaded from the registry](https://docs.npmjs.com/updating-packages-downloaded-from-the-registry)
+
+### Using lock files
+
+The lock files are a snapshot of your full dependency tree, incl. all sub-dependencies, their precise versions and also integrity hashes. Having a lock file allows npm to recreate the dependency tree exactly as it was. npm supports two lock files: `package-lock.json` and `npm-shrinkwrap.json`. Yarn uses its own lock file format.
```
ljharb, line 52:
```suggestion
Lockfiles are a snapshot of your full dependency tree, incl. all sub-dependencies, their precise versions and also integrity hashes. Having a lockfile allows npm to recreate the dependency tree exactly as it was. npm supports three lockfiles: `package-lock.json` and `npm-shrinkwrap.json`, and in npm 7+, `yarn.lock`, Yarn's format.
```

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,93 @@
...
+Accepting _all_ releases (i.e. `*` or `latest` instead of a version range, or an unbounded version range using `>`/`>=`) is risky and almost certainly will break your package unexpectedly when a new major version of the dependency is released.
+
+As with adding dependencies in general, updating them without any review carries security risks. In sensitive contexts you might want to consider gatekeeping every update by pinning the dependencies to their exact version.
+
+Further reading:
+
+- npm documentation: [About semantic version](https://docs.npmjs.com/about-semantic-versioning)
+- npm documentation: [Updating packages downloaded from the registry](https://docs.npmjs.com/updating-packages-downloaded-from-the-registry)
+
+### Using lock files
```
ljharb, line 50:
```suggestion
### Using lockfiles
```

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,93 @@
...
+
+The lock files are a snapshot of your full dependency tree, incl. all sub-dependencies, their precise versions and also integrity hashes. Having a lock file allows npm to recreate the dependency tree exactly as it was. npm supports two lock files: `package-lock.json` and `npm-shrinkwrap.json`. Yarn uses its own lock file format.
+
+The `package-lock.json` will get created automatically by default and it will not be published into the registry, i.e. it is only intended to be used for development purposes. npm recommends you commit this file into your source control.
```
ljharb, line 54:
```suggestion
`package-lock.json` will get created automatically by default and it will not be published into the registry, i.e. it is only intended to be used for development purposes. npm recommends you commit this file into your source control.
```

docs/drafts/dependency-management-guidelines.md
ljharb: i'd still like this suggestion accepted :-)

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,93 @@
...
+
+You can chose to use the `npm shrinkwrap` command to instead create an `npm-shrinkwrap.json` file. It is the same in structure as the `package-lock.json`, but it will also be published together with your package in the registry, which means that not only people who develop the package, but also the users will receive the exact set of dependencies.
+
+Using the lock files has downsides:
```
ljharb, line 58:
```suggestion
Using lockfiles has downsides:
```

docs/drafts/dependency-management-guidelines.md
```diff
@@ -0,0 +1,93 @@
...
+
+- `package-lock.json` (and `yarn.lock`) is only used in development, so you may be testing your package with a different set of dependency versions than your users will be using it, which may occasionally lead to unexpected results.
+- no dependency updates will be received automatically with `npm install`.
+- using a shrinkwrap in a package intended to be used as a library may prevent deduplication of dependencies, increasing the `node_modules` / bundle size.
```
ljharb, line 62:
```suggestion
- using `npm-shrinkwrap.json` in a package intended to be used as a library may prevent deduplication of dependencies, increasing `node_modules` / bundle size.
```

docs/drafts/dependency-management-guidelines.md
ljharb:
```suggestion
You can disable lockfiles in your global or local `.npmrc` file by adding a `package-lock=false` line. This does however mean that npm will also ignore the `npm-shrinkwrap.json` files included in your dependencies.
```

docs/drafts/dependency-management-guidelines.md
ljharb: this still needs updating

ljharb (approved):
LGTM pending remaining updates.

dominykas:
Looks like at some point I decided to delete my fork and so now this is showing "unknown repository" as source.

I restored the `dep-mgmt` branch into `nodejs/package-maintenace` and I'll open a new PR with just the changes requested, to make it easier to track the delta.

docs/drafts/dependency-management-guidelines.md
dominykas: https://github.com/nodejs/package-maintenance/pull/425/commits/e2c31a0a38e348483ab781e92added03bf6af713

docs/drafts/dependency-management-guidelines.md
dominykas: https://github.com/nodejs/package-maintenance/pull/425/commits/e2c31a0a38e348483ab781e92added03bf6af713

docs/drafts/dependency-management-guidelines.md
dominykas: https://github.com/nodejs/package-maintenance/pull/425/commits/e2c31a0a38e348483ab781e92added03bf6af713

docs/drafts/dependency-management-guidelines.md
dominykas: https://github.com/nodejs/package-maintenance/pull/425/commits/e2c31a0a38e348483ab781e92added03bf6af713

docs/drafts/dependency-management-guidelines.md
dominykas: https://github.com/nodejs/package-maintenance/pull/425/commits/e2c31a0a38e348483ab781e92added03bf6af713

docs/drafts/dependency-management-guidelines.md
dominykas: https://github.com/nodejs/package-maintenance/pull/425/commits/e2c31a0a38e348483ab781e92added03bf6af713

docs/drafts/dependency-management-guidelines.md
dominykas: https://github.com/nodejs/package-maintenance/pull/425/commits/e2c31a0a38e348483ab781e92added03bf6af713

docs/drafts/dependency-management-guidelines.md
dominykas: Remove the whole section.

thescientist13:
@dominykas 
Any thoughts on https://github.com/nodejs/package-maintenance/pull/302#pullrequestreview-513084324?

dominykas:
> Any thoughts on #302 (review)?

Sure, but can we do it after everything is merged, because https://github.com/nodejs/package-maintenance/pull/302#issuecomment-716416974 and #425?

mhdawson (approved):
LGTM

thescientist13 (approved):
Let's handle the file move in #425

lirantal (approved):
LGTM!

BethGriggs (approved):

PR #304, opened 2020-01-06
chore: removed js file

Eomm:
Deleting an unexpected js file


mhdawson:
Looks like we have enough approvals and enough time has passed, landing

Issue #373, opened 2020-06-16
draft documentation promotion


thescientist13:
We have a number of docs in [draft](https://github.com/nodejs/package-maintenance/tree/master/docs/drafts) status, but not sure what is left or needed.  They should probably all be reviewed and if complete "enough" should probably be at least moved into the top level docs folder.

This is a top level tracking item to promote them all
- [ ] Continuous Integration / Delivery (**assigned**) - #340
- [x] [Security (**assigned**)](https://github.com/nodejs/package-maintenance/blob/master/docs/drafts/security-guidelines.md) - #343
- [x] [Testing](https://github.com/nodejs/package-maintenance/blob/master/docs/drafts/testing-guidelines.md) - #344
- [x] [Publishing Guidelines (**assigned**)](https://github.com/nodejs/package-maintenance/blob/master/docs/drafts/publish-guidelines.md) - #345
- [x] [Package Support](https://github.com/nodejs/package-maintenance/blob/master/docs/drafts/package-support.md) - #346
- [x] ~~[Tools](https://github.com/nodejs/package-maintenance/blob/master/docs/drafts/tools-guidelines.md)~~ - #347 (marked as **invalid**)
- [x] Dependency Management (**assigned**) - #280
- [x] [Deprecate Guidelines](https://github.com/nodejs/package-maintenance/blob/main/docs/drafts/DEPRECATE-GUIDELINES.md) - #431 (draft) - https://github.com/nodejs/package-maintenance/pull/487

Note: they should also have the "draft" status removed from their listing in the [Table of Contents](https://github.com/nodejs/package-maintenance/tree/master/docs).

ghinks:
@thescientist13 thank you Owen, this does seem like something that I have time for. I'll look at the testing draft by @Eomm over the next week, make comments and get back.

ghinks:
@thescientist13 I think the testing document is going well and we are only waiting on a few more reviews. It is a basic document but there are quite distinct schools of thought about testing.  
I would like to move onto the CI/CD document. @dominykas mentioned in the testing document review that we need real world methodologies of how and where we can do CI/CD for open source. I agree and I'll link the testing document to the CI/CD document once we get it written.

For myself I do not know quite enough to start this yet and want to talk in the next package maintenance meeting. What we need are a bunch of projects in opensource that do this so we have concrete examples.

thescientist13:
[...]

PR #377, opened 2020-07-02
feat: testing guidelines

ghinks:
Update, fix and add to the current draft of the testing guidelines.

As part of [issue 373 ](https://github.com/nodejs/package-maintenance/issues/373#issuecomment-648762908) here are my updates to the testing guidelines.

It is essential we have some guidelines, although I believe that a whole thesis could be written on this one topic.


ghinks:
I think we should move it out. I have meeting later today and I’m hoping to get more contributors to go in after us. It’s never done done.

docs/drafts/testing-guidelines.md
```diff
@@ -16,7 +16,10 @@ The minimal versions you should focus are:
 * LTS (long time support)
 * Current
 
-Of course, you are freely to maintain a package that run also with older versions of Node.js that reach the "end-of-life" stage.
+Of course, you are free to maintain a package that also runs with older versions of Node.js that have reached the "end-of-life" stage.
```
ljharb, line 19: Perhaps something a bit stronger than "free to"? I'd personally prefer "encouraged to".

docs/drafts/testing-guidelines.md
```diff
@@ -16,7 +16,10 @@ The minimal versions you should focus are:
...
+
+For browser based modules and modules that are designed to work equally both client and server side it is essential to test on various
```
ljharb, line 21:
```suggestion
For browser-based modules and modules that are designed to work equally both client and server side it is essential to test on various 
```

docs/drafts/testing-guidelines.md
```diff
@@ -25,9 +28,32 @@ It is a good idea to have unit tests, coverage that matches most use cases for t
 * **integration test**: test your code with other applications dependencies
 * **acceptance test**: test your application sticks in performance, heavy load, etc..
 
-Here are some things to consider as you develop your package.
+#### Unit Testing
+The initial and fundamental start of any testing strategy should start with a unit testing strategy. For the JavaScript language
+the unit test may be the first place to catch code that will not parse. It is also the way to define the expectation for a particular
```
ljharb, line 33:
```suggestion
the unit test may be the first place to catch code that will not parse (a linter can achieve this as well). It is also the way to define the expectation for a particular
```

docs/drafts/testing-guidelines.md
```diff
@@ -25,9 +28,32 @@ It is a good idea to have unit tests, coverage that matches most use cases for t
...
+function or code block. The unit test should describe what to expect from that code when given a defined set of inputs.
+
+For a test to be considered a unit test it should consume no external services and only the code block under test should be considered part of that test. All external services that the code block under test consumes should be
```
ljharb, line 36: > only the code block under test should be considered part of that test

I don't agree with this; see https://martinfowler.com/bliki/UnitTest.html - unit tests should imo avoid mocking, and should effectively be an integration test of the intended behavior.

docs/drafts/testing-guidelines.md
```diff
@@ -25,9 +28,32 @@ It is a good idea to have unit tests, coverage that matches most use cases for t
...
+stubbed. There are many mocking tools available and this document will not favor one set of tools over any other.
+
+It is generally considered good practice to run units as part of any commit or build strategy.
+
+#### Integration Testing
+Integration testing in general requires that the code under test be in a package.
+
+For many packages integration testing requires that the built package be run via a testing tool upon several environments. The 
+field of continuous integration and deployment (CICD) is complex. The package being built shall be run through
```
ljharb, line 45:
```suggestion
field of continuous integration and deployment (CI/CD) is complex. The package being built shall be run through
```

docs/drafts/testing-guidelines.md
```diff
@@ -25,9 +28,32 @@ It is a good idea to have unit tests, coverage that matches most use cases for t
...
+a CICD pipeline. Several integrations are available for GitHub based repositories. (There are also alternatives to GitHub.)
```
ljharb, line 46:
```suggestion
a CI/CD pipeline. Several integrations are available for GitHub based repositories. (There are also alternatives to GitHub.)
```

docs/drafts/testing-guidelines.md
ghinks: agreed.

docs/drafts/testing-guidelines.md
ghinks: Agreed, this can be an area of contention. If I make the text clearer to indicate that the intent is to avoid the use of services such as DBs and active end points I believe it would clarify the intent I had here.

ghinks: Altered accordingly.

docs/drafts/testing-guidelines.md
```diff
@@ -25,9 +28,31 @@ It is a good idea to have unit tests, coverage that matches most use cases for t
 * **integration test**: test your code with other applications dependencies
 * **acceptance test**: test your application sticks in performance, heavy load, etc..
 
-Here are some things to consider as you develop your package.
+#### Unit Testing
```
dominykas, line 31: Should we not just link to something like https://martinfowler.com/bliki/TestPyramid.html?

docs/drafts/testing-guidelines.md
```diff
@@ -25,9 +28,31 @@ It is a good idea to have unit tests, coverage that matches most use cases for t
...
+For a test to be considered a unit test it should consume no remote services.
+
+It is generally considered good practice to run units as part of any commit or build strategy.
+
+#### Integration Testing
+Integration testing in general requires that the code under test be in a package.
+
+For many packages integration testing requires that the built package be run via a testing tool upon several environments. The /
+field of continuous integration and deployment (CICD) is complex. The package being built shall be run through
+a CICD pipeline. Several integrations are available for GitHub based repositories. (There are also alternatives to GitHub.)
```
dominykas, line 45: Pet peeve, but CI/CD pipeline is so incredibly vague, that it can mean pretty much anything... While I'm not sure we need to include these details at all (nor that the definitions of unit vs integration matter that much at all, fwiw), I think we should still link to some examples or in depth articles on this.

I'd very much be in favor of how this is done in JS world, though! Possibly with a distinction on how you do certain things in node vs browser.

ghinks: Agreed. We even have the whole document on [CI/CD](https://github.com/nodejs/package-maintenance/blob/master/docs/drafts/ci-cd-guidelines.md) dedicated to this subject. 

For the testing document maybe we should reference our CI/CD draft. 

What both yourself and @bnb have contributed to this document too.

I think we do need to have CI/CD Options documented.
- what can we do for free for an open source project
- what providers are there? 
- what does each give us?

This is actually a pretty big area in both open source and the enterprise and I think we should talk about this as an item on 14th July in the Package Maintenance meeting.

docs/drafts/testing-guidelines.md
ljharb, line 19:
```suggestion
Of course, you are encouraged to maintain a package that also runs with older versions of Node.js that have reached the "end-of-life" stage.
```

shadowspawn: I think "encouraged to maintain" comes across as too strong in that sentence, and suggests an expansive maintenance policy rather than an appropriate testing policy. The context here is what versions are being tested, and testing a range? So perhaps:

Of course, you are encouraged to also test against the older versions of Node.js your package supports.

ljharb: I'm suggesting that the expansive policy is the appropriate one. Only covering LTS is a minimum, not the ideal.

shadowspawn: I'll give a long reply since the expansive policy was deliberate, and caught my attention as something I question as appropriate in the Testing Guidelines context.

I think expansive is overly aspirational for maintainers. Encouraging maintainers to support a version of node that nodejs itself does not support, and will mean their package is running on an unsupported platform not getting bugs fixes or security fixes, likely with dependencies that are not getting bugs fixes or security fixes (from personal experience), seems a compromised version of "maintain" to encourage.

I did a quick search in case I was being overly pessimistic, but my first search results was:
https://medium.com/the-node-js-collection/maintainers-should-consider-following-node-js-release-schedule-ab08ed4de71a

I think coverage and pros and cons of support policies is an interesting separate topic, and don't like encouraging a particular pattern embedded in the Testing recommendations. 

(My impression from some comments in other topics is that it would be an improvement for new versions of node if more packages tested current and lts consistently? Impression from following limited number of the drafts and discussion, no evidence to quote!)

Not sure if others may agree, but passing on my comments. Thanks!

bnb: I am +1 to encouraging (heh) folks to stick with Node.js Maintenance releases being the furthest back release line we push projects to support. I agree that we shouldn't expect folks to support versions that we ourselves don't support. Doing so does also help move the ecosystem forward both in terms of folks being able to consistently use new language features *and* in terms of Node.js itself being able to move forward more easily.

ljharb: My belief differs; supporting an older version of node helps make those users' *upgrade of node itself* easier.

I'm fine with these docs not pushing for supporting EOL node versions, but I am not fine with them *discouraging* doing so in any way, including implying that it's not a big deal.

ghinks: So lots of discussion on this point. I think that the wording **encouraged to** leaves this open to the maintainer while indicating that support of more versions is better. I believe that we will not be able to resolve this insomuch as giving definitive advice because both sides of this discussion have valid fundamentals but differ. 

I would like to suggest that we compromise but from previous discussions on points of compromise it has been pointed out that we are actually supposed to be making **clear recommendations**.

So what would be a no compromise settlement?

I think if we say that support of LTS is considered the minimal but encourage as much support as is possible may satisfy both sides.

Thoughts?

ljharb: I'm very content with that.

shadowspawn: Short version: What about "Supporting older versions of node is encouraged where practical."?

-----

Long version

Thanks @ghinks, and I appreciate the explanation of trying for a **clear recommendation** rather than compromising early.

If "as much support as is possible" is the intended wording then my concerns still apply, that not only does this encourage a policy which is somewhat contentious, but also is not realistic for many maintainers. I am personally more concerned with the second issue, so exploring that...

A long suggestion, somewhat tongue in cheek:
"Supporting older versions of node is encouraged for the convenience of your package consumers who run older versions of node, but may not be practical due to restricting what platform features you can use, the support policies of your own dependencies, or the overheads of maintaining multiple streams of development."

This collapses down to something which might be acceptable to more people:
"Supporting older versions of node is encouraged where practical.".

The main issue that caused me to change my own support policies to "fully supported on Long Term Support versions of Node" was the testing frameworks. Not all packages state their support policy (which will improve as [PACKAGE-SUPPORT](https://github.com/nodejs/package-maintenance/blob/master/docs/PACKAGE-SUPPORT.md) is adopted, but looking at `package.json` for two popular ones I have used, currently:

Jest:  `"node": ">= 10.14.2"`

Mocha: `"node": ">= 10.12.0"`

I did also find explicit info in the Mocha [PROJECT_CHARTER](https://github.com/mochajs/mocha/blob/master/PROJECT_CHARTER.md) where it states that "Use within unmaintained versions of Node.js" is "Out of Scope".

ljharb: I'm also fine with the "where practical" qualifier, since in my experience it's impractical for very few projects to actually preserve support for very old node versions :-)

You're totally right though about testing frameworks; the only one I'm aware of that won't ever force you to drop support for some of your users is tape.

bnb: I am fine with either variant proposed.

docs/drafts/testing-guidelines.md
ghinks: @dominykas are we ok referring to the CI/CD document. I agree that we need to give concrete examples of what is currently possible in that document and I intend to start that document next.

docs/drafts/testing-guidelines.md
ghinks: I am happy to add a link

dominykas: Sorry, wasn't sure - did you mean to do it in this PR or later?

ghinks:
@thescientist13 moved from drafts to docs as requested in the meeting

docs/testing-guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Why?
+Testing is awesome!
+
+Testing is useful for developing new features, refactoring with confidence, and making sure new things don't break old things. When other people rely on your code for their own applications, testing helps you make sure things don't break for more than just yourself!
+
+Testing is even more important when new maintainers take over a package. Sometimes when we work on a codebase for a long time we forget to articulate all the cognitive load to which we have become accustomed. Good test coverage can alleviate this burden.
+
+### Updates
+Node.js has a [strict release pipeline](https://nodejs.org/en/about/releases/) focused on continuous improvement and update. In order to guarantee to your users that the module you made works correctly with the newer version of Node.js, it is a good practice to run your tests across multiple Node.js versions.
```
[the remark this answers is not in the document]
bnb: > it is a good practice to run your tests across multiple Node.js versions.

I'm confused by this. The previous statement was future looking but how this is approached it's historical or at the very least in the present. Maybe this is just a prose issue, but the sentiment doesn't seem to match the previous section.

docs/testing-guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+
+This can be done developing a CI pipeline. Check out our [CI guidelines](https://github.com/nodejs/package-maintenance/blob/master/README.md).
+
+The minimal versions you should focus are:
```
bnb, line 15:
```suggestion
The minimum versions you should focus are:
```

docs/testing-guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+* LTS (long time support)
+* Current
```
bnb, line 17: How does someone find out what these are? We should provide guidance here.

docs/testing-guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+
+Supporting older versions of node is encouraged where practical.
```
bnb, line 19:
```suggestion
Supporting older versions of Node.js is encouraged where practical.
```

docs/testing-guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+
+For browser-based modules and modules that are designed to work equally both client and server side it is essential to test on various 
+versions of your target browser and to publish which environments are supported.
```
bnb, line 22: Publish where?

docs/testing-guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+
+### How?
+It is a good idea to have unit tests, coverage that matches most use cases for the module, and make sure things work in all supported environments.
+
+* **unit test**: test your code
+* **integration test**: test your code with other applications dependencies
+* **acceptance test**: test your application sticks in performance, heavy load, etc..
```
bnb, line 29: Capitalize bullets and don't use `etc..`

bnb: Also given that you have headings under this further describing what's outlined here, you should probably make the bold parts also be relative links.

docs/testing-guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+function or code block. The unit test should describe what to expect from that code when given a defined set of inputs.
+
+For a test to be considered a unit test it should consume no remote services.
+
+It is generally considered good practice to run units as part of any commit or build strategy.
+
+#### Integration Testing
+Integration testing in general requires that the code under test be in a package.
+
+For many packages integration testing requires that the built package be run via a testing tool upon several environments. The /
```
bnb, line 43:
```suggestion
For many packages, integration testing requires that the built package be run via a testing tool upon several environments. The `/`
```

bnb: not sure if this / is intentional or not?

docs/testing-guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+field of continuous integration and deployment (CI/CD) is complex. The package being built shall be run through
+a CI/CD pipeline. Several integrations are available for GitHub based repositories. (There are also alternatives to GitHub.)
```
bnb, line 45: > Several integrations are available for GitHub based repositories. (There are also alternatives to GitHub.)

we should try to substantiate this beyond just saying it exists.

docs/testing-guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+
+For many open source projects on Github free integration environments are available.
```
bnb, line 47:
```suggestion
For many open-source projects on GitHub, free integration environments are available.
```

docs/drafts/testing-guidelines.md
dominykas: Sure, we can link and extend the existing doc. I'm just not entirely sure about the term "CI/CD pipeline" - it can mean a gazillion things to different people, so possibly we should simplify the wording towards something like "your tests should run automatically on pr/merge/commit/etc" and link to the doc?

It should not matter if it's a "pipeline" or a single job on Travis. It does not need to be complex. People need not be scared of testing and automation. The simpler the better, esp in the context of oss node packages.

docs/testing-guidelines.md
ljharb: i rarely see "open source" hyphenated; is that more common in your experience?

bnb: I go back and forth on this and try to reference external source when answering it for myself.

Here's some sources that point to it being hyphenated:
- Wikipedia hyphenates: https://en.m.wikipedia.org/wiki/Open-source_software
- Microsoft Style Guide (didn't even know this existed until searching for an answer to this now): https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/o/open-source

References that counter the change:
- Google: https://opensource.google/
- Wikipedia (lmao): https://en.m.wikipedia.org/wiki/Open_source
- GitHub: https://github.com/open-source

ljharb: given that https://opensource.org doesn't hyphenate, that suggests we shouldn't either?

docs/testing-guidelines.md
ghinks: Agreed. Do we want more alternatives for Code like GitLabs or both code and CI/CD ?

docs/testing-guidelines.md
ghinks: opps, not sure how the slash got there. Will remove.

docs/drafts/testing-guidelines.md
ghinks: I will add a link later. If that is ok. This document is going to be revisited after I do the CI/CD document that this document will link to aswell.

docs/testing-guidelines.md
ghinks: agreed, there should be a link here.

docs/drafts/testing-guidelines.md
ghinks: agreed.

docs/testing-guidelines.md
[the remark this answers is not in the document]
ghinks: After re-reading this after a break of a few days I agree with you @bnb I think some changes are required here. The first sentence about node's strict release schedule should actually be removed and some re-phrasing after wards will be required.

ghinks:
I have made a number of changes to address comments and am asking for a further review. I am acknowledging that in a number of places we do need to link to our CI/CD document. But this is a chicken and egg relationship and both are required. It is my intent to come back here to make changes once we complete the CI/CD document.

docs/testing-guidelines.md
```diff
@@ -0,0 +1,73 @@
...
+#### Integration Testing
+Integration testing in general requires that the code under test be in a package.
+
+For many packages integration testing requires that the built package be run via a testing tool upon several environments. The 
+field of continuous integration and deployment (CI/CD) is complex. The package being built shall be run through
+a CI/CD pipeline. Several integrations are available for GitHub based repositories. (There are also alternatives to GitHub.)
+These include but are not limited to 
+- [GitHub Actions](https://docs.github.com/en/actions)
+- [GitLabs](https://about.gitlab.com/)
+- [CircleCI](https://docs.github.com/en/actions)
```
rodion-arr, line 49: Looks like wrong URL
```suggestion
- [CircleCI](https://circleci.com/product/)
```

docs/testing-guidelines.md
bnb: IMO both code and CI/CD.

docs/testing-guidelines.md
ghinks: @bnb I think that we have satisfied your questions, could we have another review.

docs/testing-guidelines.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Why?
+Testing is awesome!
+
+Testing is useful for developing new features, refactoring with confidence, and making sure new things don't break old things. When other people rely on your code for their own applications, testing helps you make sure things don't break for more than just yourself!
+
+Testing is even more important when new maintainers take over a package. Sometimes when we work on a codebase for a long time we forget to articulate all the cognitive load to which we have become accustomed. Good test coverage can alleviate this burden.
+
+### Updates
+In order to guarantee to your users that the module you made works correctly with the newer version of Node.js, it is a good practice to run your tests across multiple Node.js versions.
```
bnb, line 11: > In order to guarantee

Is it a guarantee? May want to make a slightly less strong assertion.

ghinks: Agreed. I will lower the promise threshold here as we cannot offer guarantees.

ghinks: I shall be going with 
```
In order to indicate to your users that the module you made is intended to work correctly with the newer version of Node.js, it is a good practice to run your tests across multiple Node.js versions.
```

thescientist13:
@ghinks 
Oops, just noticed you have a conflict in the README.  🙈

ghinks:
I got it.

thescientist13:
I think this almost good to go now?  Any final / outstanding feedback?  @bnb do the latest changes look good to you?

thescientist13:
Discussed in the Package Maintenance meeting on the 8/25/2020 and there was consensus that we should land since there had not be a response from the originator's last objection and we believe the objection had been addressed.

bnb:
Apologies for the delay. Glad to see it landed ❤️

PR #387, opened 2020-07-20
Fix link to PUBLISH-GUIDELINES.md

rodion-arr:


thescientist13:
Good catch again @rodion-arr !

Issue #547, opened 2022-12-11
Is Conventional Package Management via Package.json still the right thing?


lemanschik:
## Discussion
Is usage of a JSON file as application entrypoint as also Installer Meta data Provider the right thing?

Overall my observation is that even the NodeJS Stack does not use the Package.json because the format is simply not optimal 
most people are switching to yarn lock for good reason and NodeJS internal already produces package-json.lock

that shows that the source format is useless anyway we could directly have a lazy Loading ESM Entrypoint. This pattern would also be more easy port able to the web.

In fact the usage and existing of a .json entrypoint broke a lot of parts of the web interop so people started to move bundlers and infrastructure to the web only to process code that is written for a so called Web Compatible Runtime and that is not logical

at last that is my observation we should stop to produce stuff that blocks other stuff and start all working like the linux community together to improve browser and whole Web Interop if NPM and NodeJS not soon gets a part of that and not the main blocker i see no future for that.

we should not reinvent the wheel and in the browser is simply no JSON import that's the current state and adding that will make nothing better it will even make it more bad. Overall you can do only one thing with Data it is ruin it.

so lets stop ruin data and create useful ways to run software under today assumptions.

ljharb:
Yes, it is.

Your comment has a bunch of statements i'd at worst call inaccurate, at best unevidenced:
 - "most people are switching to yarn lock" can you cite your source for this?
 - "with good reason" what reasons are those? either way, yarn.lock is a *lockfile* that only works in concert with package.json, so it seems irrelevant.
 - "NodeJS internal already produces package-json.lock" no, it absolutely does not; npm produces `package-lock.json` and this is also a lockfile that can only function in the presence of `package.json`
 - You can already have a lazy loading ESM entrypoint, using `import()` - in basically every environment.
 - web interop was never a thing before node, and bundlers are how we *got* web interop.
 - browsers do, in fact, have JSON import via import assertions, so this is incorrect as well.
 
To be frank, all of this seems like you have a confusingly explained }personal vendetta/agenda. The purpose of this working group is to make it easier to maintain npm packages for use in node. Specifically, and from the readme:
> Repository for discussion on how to help ensure baseline maintenance and ability to safely use key packages in the ecosystem with current Node.js versions.

I think this should be closed, and I'd encourage you to take this passion and figure out the appropriate venues to discuss it, so you can be more effective. I unfortunately don't have any suggestions besides the WinterCG group, which you're already aware of.

[...]

Issue #542, opened 2022-10-12
Create doc on how to support ESM and CommonJS when publishing modules


mhdawson:
From discussion at the last collaborator summit. 

This would help as the suggestion was that there were some easy ways to do this.

lemanschik:
@ljharb for example that you should prefer only namedExports from esm and transpile to CJS there are a lot of clashes when it comes to typescript because it ads a extra resolve matrix that is not compat

a great example are .d.ts files which use export = Module all that can get avoided if we do only namedExports by default end avoid assigning anything else then string propertys to a Object on module.exports

so never do module.exports = function. that also saves a lot of tooling quircs i can write whole books about this cross issues that come up as soon as you bundle more then 1 single code base. 

at present for example i am struggling on puppeteer and devtools-protocol also vscode makes a lot of trouble in general all the big code bases that have typescript in it are relativ blocked from ESM migration.

ljharb:
@lemanschik i strongly disagree with that; a default export is what a module *is*; named exports are what module *has*, and most modules should *be* something. That’s a stylistic choice and not something node should be weighing in on. I, and the Airbnb style guide ftr, say that you should almost *always* do module.exports = function.

lemanschik:
@ljharb they say so because of compat i am working since ages with ESM only and can report back that no one did produce correct ESM Builds that load in a correct ESM Environment. (Other then NodeJS)

ljharb:
I have a number of projects that work perfectly compatibly with a build process, so I’m not sure about those reports.

lemanschik:
@ljharb your again correct with build tools which correct the imports everything works at the end but the Nativ code that should work untranspiled and should even get patched at runtime does not work. 

thats why i got rollup and typescript contributor only because of such issues.

lemanschik:
Maybe it helps to show a true isomorphic cjs ESM module that uses a global 

Load Isomorphic CJS that uses only dynamic import and assigns to global as ESM and CJS share the same global this works. 

```js
globalThis.module = { exports: {}}
import('').then(_m=>(v => {
 const v = globalThis.module.exports
 delete globalThis.module.exports
 return v
});
```

lemanschik:
This pattern is also nice when you work with CJS ESM NodeJS


my_esm.js
```js
const moduleName = { fnName() {} }
export { moduleName }
```

my_cjs_wrapper
```js

const moduleName = {
   async fnName() {
    return (await(import('./my_esm.js')).fnName()
[...]

PR #548, opened 2022-12-11
Need to leave that effort in favor of net wide nodejs deprecation via…

lemanschik:
… chromium modular as a platform it self

Need to leave that effort in favor of net wide nodejs deprecation via chromium modular as a platform it self

see: 
- https://github.com/nodejs/package-maintenance/issues/547#event-8006766334
- https://github.com/nodejs/package-maintenance/issues/542#issuecomment-1345345127

It makes out of my view no sense to keep NPM and NodeJS Compat when the community is blocking so i need to deprecate nodejs via offering a chromium cli which i in fact have at hands to replace nodejs via a modular chromium build it self i mainly work on the chromium project anyway.


ljharb (approved):
good luck to you

Issue #517, opened 2022-03-10
Update `node` namespace values in package support


bnb:
Hey!

For a project, I was trying to implement the node namespace values.  I took some notes as I was exploring it, which you can find raw [here](https://gist.github.com/bnb/d15a8ff30c60f7c3cd47295932a64d52). However, I'd like to refine them and share my issues/concerns with the `node` namespace values:

- Currently, there is inconsistency in how the values are named.
  - `lts` is equivalent to `lts_all` if we're looking at the naming of both `all` and `lts_latest`/`lts_active`. We have two naming schemes side by side - effectively, we've implemented both a flat structure and a tiered structure, and the intersection of those two is messy.
- `active` is particularly problematic since it co-opts existing terminology.
  - `Active` is terminology already adopted by the release working group for LTS releases. `active` in the context of this includes both (Release WG terminology) `current` and `active LTS`, which does not help clarify the already confusing state that we exist within.
- Certain matrix combinations are missing.
  - Example: there's not a way for someone to say they only support maintenance LTS.
  - Example: there's not a way for someone to say they only support EOL versions.
  - While these examples may not be something we'd immediately consider as valid options, there are potentially scenarios where they'd be useful. For example, imagine a polyfill library that backfills all possible new Node.js features to maintenance mode LTS, or a package that intentionally logs a message on EOL verisons.

Rather than building out an incomplete structure (especially when we've collectively consistently struggled with naming release concepts in the past!) as a matrix, I'd like to propose that we simply take the elements of the matrix and directly support those as an array rather than a string.

  - current
  - lts/latest (alternate: `lts-latest`)
  - lts/active (alternate: `lts-active`)
  - lts/maintenance ((alternate: `lts-maintenance`)
  - eol

any combination of these achieves every option in the existing namespace:

- all: `current, lts/active, lts/maintenance, eol`
- lts: `lts/active, lts,maintenance`
- lts_latest: `lts/latest`
- supported: `current, lts/active, lts/maintenance`
- current: `current`

the _only_ variation from the Release WG definition of release lines is `lts/latest` which is a subset that's independent of categories, rather than a combination of multiple categories.

bnb:
I've published a module that accomplishes providing full information (whereas nv only currently implements a partial dataset) of the proposed properties: http://npm.im/@nodevu/ranges

mhdawson:
[...]

PR #581, opened 2023-10-26
doc: update `node` namespace values

bnb:
This PR updates the `node` namespace values. It closes https://github.com/nodejs/package-maintenance/issues/517.

### DIFF
--- a/docs/PACKAGE-SUPPORT.md
+++ b/docs/PACKAGE-SUPPORT.md
@@ -206,22 +206,25 @@ The standardized namespace identifiers are:
 
 The standardized options for the `node` namespace are as follows:
 
-| Value                | Current | Active LTS | Maintenance LTS | EOL | Example    | Description |
-|----------------------|---------|------------|-----------------|-----|------------|-------------|
-| (valid semver range) |         |            |                 |     |             | A [semver range](https://semver.io/) of Node.js versions supported.
-| `abandoned`          |         |            |                 |     |             | Not recommended for use. The package is deprecated or no longer maintained
-| `none`               |         |            |                 |     |             | Use at your own risk, no active support. May or may not work for a given Node.js version
-| `all`                |   ✔     |     ✔      |        ✔        |  ✔  | ...11,12,13,14,15 | The package is maintained for versions of Node.js including both LTS and non-LTS releases regardless of whether they are EOL or not. It may be necessary to accept semver-major level (ie. breaking) changes into that application in order to receive essential fixes. Documentation for the package will include the non-LTS releases for which the package is still maintained (some maintainers support as far back as 0.6)
-| `lts`                |         |     ✔      |        ✔        |     | 10,12       | The package is maintained for the Node.js LTS releases (both in Active and Maintenance status). Anyone creating an application using an LTS version of Node.js and using the latest major version of LTS adopting packages will not have to accept semver-major level (ie. breaking) changes into that application in order to receive essential fixes. Full details are available [here](https://github.com/CloudNativeJS/ModuleLTS)
-| `active`             |   ✔     |     ✔      |                 |     | 10,12,14,15 | Current release and active LTS releases
-| `lts_active`         |         |  ✔ (all)   |                 |     | 10,12       | All releases in _active_ LTS status. There may be more than one LTS release in active maintenance at a given point in time
-| `lts_latest`         |         | ✔ (latest) |                 |     | 12          | The package is maintained only for the Latest LTS Node.js version. You will be required to update to the latest LTS Node.js version in order to ensure you can use new versions/get security fixes
-| `supported`          |   ✔     |     ✔      |        ✔        |     | 10,12,14,15 | Node.js versions which are not EOL
-| `current`            |   ✔     |            |                 |     | 15          | The latest release from "all"
+| name | alias | description |
+|------|-------|-------------|
+| current |  | The Node.js release that is defined to be "Current" status by the [Node.js Release Process](https://github.com/nodejs/release). |
+| lts/latest | lts-latest | A subset of the Node.js LTS releases, only including the the _most recent_ Node.js Release that is in "Active" LTS status as defined by the [Node.js Release Process]( https://github.com/nodejs/release). |
+| lts/active | lts-active | A subset of the Node.js LTS releases, including _all_ of the Node.js Releases that are in "Active" LTS status as defined by the [Node.js Release Process](https://github.com/nodejs/release). |
+| lts/maintenance | lts-maintenance | A subset of the Node.js LTS releases, including _all_ of the Node.js Releases that are in "Maintenance" LTS status as defined by the [Node.js Release Process](https://github.com/nodejs/release). |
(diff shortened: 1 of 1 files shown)


docs/PACKAGE-SUPPORT.md
```diff
@@ -206,23 +206,26 @@ The standardized namespace identifiers are:
...
-| `supported`          |   ✔     |     ✔      |        ✔        |     | 10,12,14,15 | Node.js versions which are not EOL
-| `current`            |   ✔     |            |                 |     | 15          | The latest release from "all"
+| name | alias | description |
+|------|-------|-------------|
+| current |  | The Node.js release that is defined to be "Current" status by the [Node.js Release Process](https://github.com/nodejs/release). |
+| lts/latest | lts-latest | A subset of the Node.js LTS releases, only including the the _most recent_ Node.js Release that is in "Active" LTS status as defined by the [Node.js Release Process]( https://github.com/nodejs/release). |
+| lts/active | lts-active | A subset of the Node.js LTS releases, including _all_ of the Node.js Releases that are in "Active" LTS status as defined by the [Node.js Release Process](https://github.com/nodejs/release). |
+| lts/maintenance | lts-maintenance | A subset of the Node.js LTS releases, including _all_ of the Node.js Releases that are in "Maintenance" LTS status as defined by the [Node.js Release Process](https://github.com/nodejs/release). |
```
ljharb, line 214: where does, for example, v18.18.1 fall? It's active LTS, but now v18.18.2 has precluded it. Does "lts/active" include all versions that were ever LTS in an actively maintained line? (meaning, the lower threshold only becomes a breaking change when an LTS line goes EOL)

bnb: What is the status quo? My guess is that it would be contextual. For support, it would be the full range. For something like CI, I would guess it'd be most recent in-range.

wesleytodd: I don't think we established an official status quo here. I think that we might need `lts-maintenence` and `maintenance` or something similar for this use case. Or maybe do it the other way around? `lts-maintenance` and `lts-maintenance-all`?

wesleytodd: As a point of reference, the `@pkgjs/nv` tool only returns the latest for each major when asked for `lts_active` currently. This has been a point of issue for me, and so just added `latestOfMajorOnly` but didn't add it to the cli flags.

docs/PACKAGE-SUPPORT.md
```diff
@@ -206,23 +206,26 @@ The standardized namespace identifiers are:
...
-| `active`             |   ✔     |     ✔      |                 |     | 10,12,14,15 | Current release and active LTS releases
-| `lts_active`         |         |  ✔ (all)   |                 |     | 10,12       | All releases in _active_ LTS status. There may be more than one LTS release in active maintenance at a given point in time
-| `lts_latest`         |         | ✔ (latest) |                 |     | 12          | The package is maintained only for the Latest LTS Node.js version. You will be required to update to the latest LTS Node.js version in order to ensure you can use new versions/get security fixes
-| `supported`          |   ✔     |     ✔      |        ✔        |     | 10,12,14,15 | Node.js versions which are not EOL
-| `current`            |   ✔     |            |                 |     | 15          | The latest release from "all"
+| name | alias | description |
+|------|-------|-------------|
+| current |  | The Node.js release that is defined to be "Current" status by the [Node.js Release Process](https://github.com/nodejs/release). |
```
wesleytodd, line 211: There is precidence in other areas for calling this `latest` (the npm registry), should we align on that? We could ask the build folks to start calling it `latest`?

docs/PACKAGE-SUPPORT.md
```diff
@@ -206,23 +206,26 @@ The standardized namespace identifiers are:
...
+| lts/latest | lts-latest | A subset of the Node.js LTS releases, only including the the _most recent_ Node.js Release that is in "Active" LTS status as defined by the [Node.js Release Process]( https://github.com/nodejs/release). |
```
wesleytodd, line 212: This implies that it can be multiple then clarifies that it is a set of one. I am not sure if this language is helpful. I think we should say it is either one release or multiple.

docs/PACKAGE-SUPPORT.md
```diff
@@ -206,23 +206,26 @@ The standardized namespace identifiers are:
...
+| `lts` | `lts/active, lts/maintenance` | The package is maintained for the Node.js LTS releases (both in Active and Maintenance status). Anyone creating an application using an LTS version of Node.js and using the latest major version of LTS adopting packages will not have to accept semver-major level (ie. breaking) changes into that application in order to receive essential fixes. Full details are available [here](https://github.com/CloudNativeJS/ModuleLTS) |
+| `lts_latest` | `lts/latest` | The package is maintained only for the Latest LTS Node.js version. You will be required to update to the latest LTS Node.js version in order to ensure you can use new versions/get security fixes |
+| `supported` | `current, lts/active, lts/maintenance` | Node.js versions which are not EOL |
+| `current` | `current` | The latest release from "all" |
```
wesleytodd, line 225: FWIW, I think the old vs translated here is confusing in some examples. I think partly that is because the new ones cover more nuanced meaning and for example `all` as the old one is not even present in the new context. I honestly would rather just drop the "old description" and replace that with "whats changed" or something?

wesleytodd (changes requested):
From the one discussion, I think we need specifiers for "latest of the major line" which are separate from "all versions in the group" before we move forward with this PR. The rest of my comments are more nitpicks and should not be considered blocking in any way.

bnb:
> From the one discussion, I think we need specifiers for "latest of the major line" which are separate from "all versions in the group" before we move forward with this PR. The rest of my comments are more nitpicks and should not be considered blocking in any way.

This point is a distinct issue with how this was initially implemented, not with this PR. This PR doesn't change the status quo - that should be done in a different PR, as it's a different set of work. I'd be happy to work on that separately, but this PR should IMO be judged on how it stacks against the status quo and the gaps in that should not be the responsibility of the PR to fix.

wesleytodd:
Hm, I think I disagree because even though this is not a code repo, this is a breaking change and so would be the change to address this gap. Since "publishing" here is just merging this, I would prefer if we landed all the related breaking changes in one go. I guess I am good with doing it in two PR's where we block both until they are both ready.

That said, I don't think this is important enough to block progress and I am fully in support of these changes, so I will withdraw my blocking review/comment.

wesleytodd (approved):
Hm, I think I disagree because even though this is not a code repo, this is a breaking change and so would be the change to address this gap. Since "publishing" here is just merging this, I would prefer if we landed all the related breaking changes in one go. I guess I am good with doing it in two PR's where we block both until they are both ready.

That said, I don't think this is important enough to block progress and I am fully in support of these changes, so I will withdraw my blocking review/comment.

docs/PACKAGE-SUPPORT.md
shadowspawn: There is a line below the table which states: "Unless the package documentation explicitly states otherwise, the support is only implied for the latest release in any major release line."

wesleytodd: Now that you mention it, I remember us discussing this once in the original work on this. I still think we need to be much more explicit with these aliases meaning around this point.

PR #592, opened 2024-03-28
Add Charter

aduh95:
Not sure if we actually need that, or if this is the right approach. The idea would be to delegate the discussions from TSC meetings to a different group.

### DIFF
--- a/Governance.md
+++ b/Governance.md
@@ -41,7 +41,7 @@ The WG has authority over Version Management solutions in the project including:
 
 [^note]A note on the meaning of "maintaining the relationship":
 
-This means proposals to change the relationship between Node.js and npm or other tools (from either side) should be presented to the PMWG first to work out the details of the proposed changed.
+This means proposals to change the relationship between Node.js and npm or other tools (from either side) should be presented to the PMWG first to work out the details of the proposed change.
 
 ## Collaborators
 


Governance.md
```diff
@@ -1,289 +1,177 @@
...
+* Maintaining the relationship[^note] with any tools included in the Node.js default distribution with the express purpose of the following:
+  - Download or install additional software from sources outside the default distribution.
+  - Download or install additional Node.js distributions (default or third-party).
+* Contribution policy for all of the additional software included in the Node.js distribution as qualified above.
+* Conduct guidelines for the Version Management group.
+* Recommend technical solutions for Version Management for the Node.js ecosystem.
+
+[^note]A note on the meaning of "maintaining the relationship":
+
+This means proposals to change the relationship between Node.js and npm or other tools (from either side) should be presented to the PMWG first to work out the details of the proposed changed.
```
GeoffreyBooth, line 44:
```suggestion
This means proposals to change the relationship between Node.js and npm or other tools (from either side) should be presented to the PMWG first to work out the details of the proposed change.
```

Governance.md
```diff
@@ -1,289 +1,177 @@
...
+  following repositories:
+  * nodejs/ci-config-travis
+  * nodejs/ci-config-github-actions
+  * nodejs/package-maintenance
+
+For the current list of WG members, see the project [README.md][].
+
+[README.md]: ./README.md#current-project-team-members
+
+### Version Management Tools
```
wesleytodd, line 31: For folks seeing such a large diff and wondering what changed, this is the section to focus on. The rest is marked as changed but really isnt. Additionally, the parts above about the repo's is because the goal here is to subsume the corepack team into the PMWG.

wesleytodd:
Goals of this change:

1. Give discussions around package managers of all varieties a place to have dedicated discussions around how they interop with Node.js (both the technical and the non-technical)
2. Avoid Node.js Core PRs being the only avenue for having active discussion on these topics (avoid battle pr's https://github.com/nodejs/node/pull/51981 and https://github.com/nodejs/node/pull/51886 for example)
3. Help give more resources to @GeoffreyBooth's work to help define Node.js policy around this stuff
4. Make it unambiguous where to seek consensus for future discussions before raising related topics to Node.js core or TSC

Governance.md
```diff
@@ -1,289 +1,177 @@
-## Package Maintenance Working Group Governance
+## Package Maintenance Working Group Charter
+
+The Node.js Package Maintenance Working Group is jointly governed by the members
+of that Working Group (WG), and maintains oversight of its constituting teams:
+the `@nodejs/package-maintenance` and `@nodejs/corepack` teams.
+
+The WG responsibilities include:
+
+* Management of the "version management tools" shipped with the Node.js distribution (see below).
```
GeoffreyBooth, line 9: I’m not sure whether whatever potential new tool we make would necessarily ship bundled in the distribution; and of course we might unbundle Corepack. We should still be responsible for these things regardless, though, so I think a better way to put it is that we’re responsible for anything under https://github.com/nodejs.

```suggestion
* Management of the “version management tools” maintained within the Node.js GitHub org.
```

Governance.md
```diff
@@ -1,289 +1,177 @@
...
+
+For the current list of WG members, see the project [README.md][].
+
+[README.md]: ./README.md#current-project-team-members
+
+### Version Management Tools
+
+The WG has authority over Version Management solutions in the project including:
+
+* Maintaining the relationship[^note] with any tools included in the Node.js default distribution with the express purpose of the following:
```
GeoffreyBooth, line 35:
```suggestion
* Maintaining the relationship[^note] with any tools included or referenced by Node.js-managed software with the express purpose of the following:
```

Governance.md
```diff
@@ -1,289 +1,177 @@
...
+  - Download or install additional software from sources outside the default distribution.
+  - Download or install additional Node.js distributions (default or third-party).
+* Contribution policy for all of the additional software included in the Node.js distribution as qualified above.
```
GeoffreyBooth, line 38:
```suggestion
* Contribution policy for all of the additional software within the Node.js GitHub organization as qualified above.
```

Governance.md
wesleytodd: Ah interesting point. Hm, yeah I think this is a good change but npm is not maintained within this org, so that feels at odds right?

GeoffreyBooth:
In the collab summit meeting meeting on 2024-04-03 we discussed trying to agree on goals first, and getting those approved by both this team and then the TSC, and then chartering this team to work out the implementation details to achieve those goals.

Governance.md
wesleytodd: Hm interesting change. I think I understand why but would you mind explaining the goal of this wording over the other?

Issue #593, opened 2024-04-03
[Draft] Proposal for Node.js Version Management


wesleytodd:
## The History

Version Management is something the Node.js project has discussed a few times. Starting in *< INSERT HISTORY OF WG HERE >*. And more recently with the introduction of `corepack` for managing versions of package managers. While there has been a volume of discussions and code examples I do not believe there has been a cohesive technical document to bring it all together. This is my attempt at such a doc.

Node.js and the toolchain it ships with is used throughout the SDLC (software development lifecycle) and has different needs at different stages of work. For example, it is highly
uncommon for someone to use a different versions of anything within the same production deployment while is is highly common to do so for local development or CI.

This has led to a the Node.js project focusing on the well understood (but also highly complicated) aspect of simply supporting one locked version of the toolchain across the many
supported architectures. This is in fact a required area of investment for the project as no matter the local dev toolchain a team or developer chooses, when in prod it is *always*
necessary to support a secure and stable single version workflow.

This focus has left it to the ecosystem to solve for the local and CI workflows where many versions are commonly needed for a variety of tools. And when the ecosystem is in charge
a wonderful variety of solutions emerge. 

## The Problem

A realistic and productive Node.js development environment cannot be setup with *only* tools shipped within the current executables. A version manager for both the `node`
executable and a package manager is required.

## The Stakeholders

This proposal requires members of the community to work together. To help enable that, I propose we formalize the Package Maintenance Working Group (PMWG) to own this domain and be a
shared space for Node.js contributors, tooling authors, and interested community members to work. That would help ensure that all stakeholders in the ecosystem had a well
documented and clear idea of who to work with and how that process proceeds.

Once we have chartered the PMWG to take ownership of this domain we can expand on this proposal and other related document and policies around how the Node.js Project intends to
move forward with regards to Package Mangers, version managers, and associated tooling which are vendored or included in some from with the Node.js distribution.

## Use Case Scope

We cannot build all things for all people. To help keep focus within this proposal we want to focus on solving for one specific use case:

**Users installing Node.js on their local development machine or in CICD process for the purpose of building.**

This means we are not intending for this tool to be used to install on production servers. That is not only an already well covered use case, it is also not a place where
dynamically installing many runtime or tool versions is a good or recommended practice.

## Goals

1. A new solution should attempt to reduce redundancy in both concepts and metadata in package.json were possible
[...]

PR #594, opened 2024-04-04
feat: add proposal for binary management

wesleytodd:
Moved from #593. It should have been a PR but I just was doing that in the collab summit before the session and I didn't have the PMWG repo closed down at the time. Sorry about that.

### DIFF
--- a/docs/binary-management.md
+++ b/docs/binary-management.md
@@ -1,0 +1,263 @@
+# Manage all the things you need to build with Node.js
+
+## The History
+
+Version Management is something the Node.js project has discussed a few times. Starting in *< INSERT HISTORY OF WG HERE >*. And more recently with the introduction of `corepack` for managing versions of package managers. While there has been a volume of discussions and code examples I do not believe there has been a cohesive technical document to bring it all together. This is my attempt at such a doc.
+
+Node.js and the toolchain it ships with is used throughout the SDLC (software development lifecycle) and has different needs at different stages of work. For example, it is highly
+uncommon for someone to use a different versions of anything within the same production deployment while is is highly common to do so for local development or CI.
+
+This has led to a the Node.js project focusing on the well understood (but also highly complicated) aspect of simply supporting one locked version of the toolchain across the many
+supported architectures. This is in fact a required area of investment for the project as no matter the local dev toolchain a team or developer chooses, when in prod it is *always*
+necessary to support a secure and stable single version workflow.
+
+This focus has left it to the ecosystem to solve for the local and CI workflows where many versions are commonly needed for a variety of tools. And when the ecosystem is in charge
+a wonderful variety of solutions emerge. 
+
+## The Problem
+
+A realistic and productive Node.js development environment cannot be setup with *only* tools shipped within the current executables. A version manager for both the `node`
+executable and a package manager is required.
+
+## The Stakeholders
+
+This proposal requires members of the community to work together. To help enable that, I propose we formalize the Package Maintenance Working Group (PMWG) to own this domain and be a
+shared space for Node.js contributors, tooling authors, and interested community members to work. That would help ensure that all stakeholders in the ecosystem had a well
+documented and clear idea of who to work with and how that process proceeds.
+
+Once we have chartered the PMWG to take ownership of this domain we can expand on this proposal and other related document and policies around how the Node.js Project intends to
+move forward with regards to Package Mangers, version managers, and associated tooling which are vendored or included in some from with the Node.js distribution.
+
+## Use Case Scope
+
+We cannot build all things for all people. To help keep focus within this proposal we want to focus on solving for one specific use case:
+
+**Users installing Node.js on their local development machine or in CICD process for the purpose of building.**
+
+This means we are not intending for this tool to be used to install on production servers. That is not only an already well covered use case, it is also not a place where
+dynamically installing many runtime or tool versions is a good or recommended practice.
+
+## Goals
+
+1. A new solution should attempt to reduce redundancy in both concepts and metadata in package.json were possible
+2. Reduce developer toil (as @GeoffreyBooth said: The most important part of this is that it defines the end goals: no lockfile mismatches, no failed builds.)
+
+## Prior Art
+
+There is alot here, adding a place holder to remember but not get distracted.
+
+@TODO list out more about the current version managers and their approaches.
+
+## A Proposal
+
+Node.js ships or endorses (either builds or adopts an existing project) a version manager with the following qualities:
+
+1. Manage versions of both `node` and a **small** subset of ecosystem tooling for local and CI environments
+2. The tool should enhance the user experience for all supported tools in an equitable way (note on this below)
+3. There is a clear and concise process for tools to be included (but as we are not trying to achieve equality, some tools are grandfathered in)
+4. The technical solution (and all others considered) are well documented and clear to stakeholders
+5. The technical solution should be reasonably approachable for our consumers without a lot of new education required
+
+As there is already a lot of prior art in the ecosystem. I will venture to make a clear proposal for an end user experience we hope to achieve without picking an existing tool or
+set of implementation details (hence the name `nver` as a placeholder):
+
+```
+$ curl -o- https://nodejs.org/dist/install.sh | bash
+$ node --version # the latest lts version
+$ npm --version # the npm version shipped with node
+$ mkdir project && cd project
+$ npm init -y
+$ nver install npm@latest node@20.x
+$ node --version # the local node version 20.x
+$ npm --version # the local npm version newer than what was shipped with node at 20.x
+```
+
+While this is the "bootstrap from scratch UX" there are many other important workflows which led to the development of `corepack` which we should also well support. One such is
+cloning down an existing project and ensuring that you have a properly setup development environment for working with it. Here is an example for that UX:
+
+```
+$ git clone git@github.com:nodejs/project.git && cd project
+$ npm install
+This project requires packages to be installed with pnpm@8.15.4 and node@^20.0.0.
+We can help you set that up by answering yes to the prompt below, or after selecting
+no you can run:
+
+$ nver install
+
+Continue? (yes)
+...
+$ pnpm install
+```
+
+You may notice in the above that the `npm` executable noticed that the project was not meant for use with `npm` and displayed a message to users. This proposal would include some
+collaboration by our esteemed colleagues at the various supported packages managers to implement a common stub which would be maintained by the Node.js project which implements a
+standard check. The inclusion of this stub would be a requirement for being included in the tooling. To achieve this we have been discussing a proposal in the Package Metadata Interop Collab
+Space to add a `devEngines` field. More on that below.
+
+And lastly, most developers work on more than one project at a time. It is important for their productivity that it is easy to switch between projects and their associated tooling.
+Here is an example of working on multiple projects at once:
+
+```
+$ cd project1
+$ npm it # install and test (assuming project is already configured or using defaults)
+$ node --version
+v20.11.1
+$ cd ../project2
+$ npm it
+This project requires packages to be installed with yarn@1.22.21 and node@^18.0.0.
+We can help you set that up by answering yes to the prompt below, or after selecting
+no you can run:
+
+$ nver install
+
+Continue? (yes)
+...
+$ yarn install && yarn test
+$ node --version
+v18.19.1
+$ cd ../project1
+$ npm test # seamless switch back to node version
+```
+
+### Declaring development engines
+
+Currently the only widely used way to declare development tooling is the `packageManager` field. This only covers one aspect of the problem (the package manager) and is not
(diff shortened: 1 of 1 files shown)


wesleytodd:
Not sure if we need to copy this over to here: https://github.com/nodejs/package-maintenance/issues/593#issuecomment-2035146777

Closing that in favor of the PR version.

docs/binary-management.md
```diff
@@ -0,0 +1,263 @@
...
+    // runtime: '', as specified in original proposal
+    packageManager: {
+      name: 'npm', // names which we would specify and map to individual known projects
+      version: '', // this would be semantically left open from the spec, but should be considered the equivalent of npm package specifiers
+      // see: https://github.com/zkat/ssri/blob/latest/README.md#--ssriparsesri-opts%E2%80%94integrity 
+      resolved: {
+        version: '1.0.0',
+        uri: 'https://registry.url/path/to/tarball.tgz',
+        // this would be the output of ssri
+        digest: '...'
```
styfle, line 150:
```suggestion
    "packageManager": {
      "name": "npm", // names which we would specify and map to individual known projects
      "version": "", // this would be semantically left open from the spec, but should be considered the equivalent of npm package specifiers
      // see: https://github.com/zkat/ssri/blob/latest/README.md#--ssriparsesri-opts%E2%80%94integrity 
      "resolved": {
        "version": "1.0.0",
        "uri": "https://registry.url/path/to/tarball.tgz",
        // this would be the output of ssri
        "digest": "sha1.c85a4305534f76d461407b59277b954bac97b5c4"
```

docs/binary-management.md
```diff
@@ -0,0 +1,263 @@
...
+{
+  "devEngines": {
+    // os: '', as specified in original proposal
+    // cpu: '', as specified in original proposal
+    // runtime: '', as specified in original proposal
+    packageManager: {
+      name: 'npm', // names which we would specify and map to individual known projects
+      version: '', // this would be semantically left open from the spec, but should be considered the equivalent of npm package specifiers
+      // see: https://github.com/zkat/ssri/blob/latest/README.md#--ssriparsesri-opts%E2%80%94integrity 
+      resolved: {
```
GeoffreyBooth, line 146: Can we avoid defining a custom addition to the `devEngines` spec? It sounds like what you’re creating here is a lockfile; but there are already lockfiles that accompany `package.json` files. I’m not sure we need a standardized lockfile, but if we do, that seems like it should be something proposed on its own?

docs/binary-management.md
wesleytodd: I believe the output from `ssri` can be more than just this string right? I didn't want to try to make it look like we were documenting the behavior of that package with this line, but if that is the output I am good with this change as a whole.

docs/binary-management.md
wesleytodd: This is what I propose we change the `devEngines` spec change to. This is an explicit change I propose we make there.  The existing field already had changes requested in that issue which were not addressed, this attempts to address them. Specifically the name is changed to not suggest that the spec requires implementors to implement "download" functionality.

GeoffreyBooth:
This is a huge document that feels like something that would come out of a monthslong discovery process. Can we maybe mark this as draft for now and start with a new document that just defines the goals we want to address? Let’s get that written down and hopefully find consensus there, before moving on to various approaches and their pros and cons, etc.

One model we could follow is https://github.com/nodejs/loaders/tree/main/doc, where separate files cover prior art, external resources, etc. This PR could be split up into multiple files by topic and then some of them can land without waiting for the goals discussion, such as a doc listing prior art.

docs/binary-management.md
styfle: I figured you had "uri" and it could be anything so either add examples to all or none

wesleytodd: Ah yeah, I see the dissonance in the example. I guess probably what we should do is take the output from `ssri` via a `package-lock.json` as an example.

wesleytodd:
> This is a huge document that feels like something that would come out of a monthslong discovery process. Can we maybe mark this as draft for now and start with a new document that just defines the goals we want to address?

Yeah I agree if is too long/much. And I did mark it as draft after reading this comment the first time yesterday. That said, I am not clear on what we get out of multiple documents. I agree this is not complete, but I think part of the problem that was burning folks out in the previous parts of this conversation was too many separate pr's/issues. As I can totally see it is too long, I am concerned about sharding the conversation again. Would it make you more comfortable if I deleted some parts for now (like some of the solutioning I did) and then we can focus for now on the prior art and goals sections but keep it in one doc?

wesleytodd:
One other thing to note which I was just reminded of by a comment in one of this repos issues, the intent is that with this proposal the PMWG takes over the scope of the old version management effort here: https://github.com/nodejs/version-management

GeoffreyBooth:
> That said, I am not clear on what we get out of multiple documents.

It means that simple parts can just land quickly, like the list of prior art. It's also more useful to have separate files rather than one big one. If you look in the loaders repo for example, we have several design proposals in separate files. That's much easier to work with than everything in one file.

You don't need to do one PR per file. Maybe start with one PR for all the history stuff, prior art and links/resources, any other related topics that can be one file per topic. That should be easy to land without much controversy, and we can tackle goals in its own PR.

wesleytodd:
Ok, I can work later this week on breaking it up. That said, I would be happy to have comments on what is here before then because largely I have struggled to carve out the time to complete this, so when I get back to it if I can address many things at once it would be more efficient. Also, I don't need to be the person to do this work so if anyone wants to help move this along faster than I am able that would be awesome as well.

GeoffreyBooth:
I'm happy to help with the documentation work. My availability is also scattered so it'll be in dribs and drabs over the next few days/weeks.

I think the most important thing though is to start collecting a list of goals that either have consensus or majority support. Maybe we could start with the consensus ones and separately add ones that require voting as a later PR. Is the best way to do this via a PR or maybe a GitHub discussion?

docs/binary-management.md
```diff
@@ -0,0 +1,263 @@
...
+supported architectures. This is in fact a required area of investment for the project as no matter the local dev toolchain a team or developer chooses, when in prod it is *always*
+necessary to support a secure and stable single version workflow.
+
+This focus has left it to the ecosystem to solve for the local and CI workflows where many versions are commonly needed for a variety of tools. And when the ecosystem is in charge
+a wonderful variety of solutions emerge. 
+
+## The Problem
+
+A realistic and productive Node.js development environment cannot be setup with *only* tools shipped within the current executables. A version manager for both the `node`
+executable and a package manager is required.
```
GeoffreyBooth, line 20: A version manager is certainly not a requirement, either for runtime or for package manager. It's much more common for runtime, but a common workflow is to develop locally using latest Node and test for a specific locked version using Docker or a nonprod hosted environment. That's how I've seen many development teams operate over many years.

In the last decade at work I've seen maybe three dozen or so Node app development teams and none of them have used a package manager version manager. All of those teams have used either npm or a handful used Yarn 1. From what I can tell, package manager version managers are really only popular for pnpm because of its bug in pnpm 8 where the lockfile changes often.

Also this section purports to define the problem but really it's describing a solution: one or more version managers. What problems are those tools solving?

docs/binary-management.md
```diff
@@ -0,0 +1,263 @@
...
+1. A new solution should attempt to reduce redundancy in both concepts and metadata in package.json were possible
+2. Reduce developer toil (as @GeoffreyBooth said: The most important part of this is that it defines the end goals: no lockfile mismatches, no failed builds.)
+
+## Prior Art
+
+There is alot here, adding a place holder to remember but not get distracted.
+
+@TODO list out more about the current version managers and their approaches.
+
+## A Proposal
```
GeoffreyBooth, line 51: This proposal should become its own file, but I would suggest setting it aside until we agree on goals because it might change to address them.

docs/binary-management.md
GeoffreyBooth: I'm unaware of requests for changes on that spec that haven't been addressed; but if there are, let's deal with notes over there rather than defining a competing alternative.

docs/binary-management.md
```diff
@@ -0,0 +1,263 @@
...
+    }
+  }
+}
+```
+
+With this slimmed down version, we can test out approaches with different tooling implementations for how we resolve, fetch, and install the dependency without forcing the hand of
+tool authors into one single direction. If they wanted to limit that `version` cannot be a git repository, that would be acceptable even though `npm` defines that as valid for a
+package specifier.
+
+## Technical Approachs
```
GeoffreyBooth, line 161:
```suggestion
## Technical Approaches
```

docs/binary-management.md
```diff
@@ -0,0 +1,263 @@
...
+
+Version Management is something the Node.js project has discussed a few times. Starting in *< INSERT HISTORY OF WG HERE >*. And more recently with the introduction of `corepack` for managing versions of package managers. While there has been a volume of discussions and code examples I do not believe there has been a cohesive technical document to bring it all together. This is my attempt at such a doc.
+
+Node.js and the toolchain it ships with is used throughout the SDLC (software development lifecycle) and has different needs at different stages of work. For example, it is highly
+uncommon for someone to use a different versions of anything within the same production deployment while is is highly common to do so for local development or CI.
+
+This has led to a the Node.js project focusing on the well understood (but also highly complicated) aspect of simply supporting one locked version of the toolchain across the many
+supported architectures. This is in fact a required area of investment for the project as no matter the local dev toolchain a team or developer chooses, when in prod it is *always*
+necessary to support a secure and stable single version workflow.
```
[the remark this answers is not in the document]
styfle: Correct, Vercel and AWS only support major version. Minor and patch are automatically upgraded.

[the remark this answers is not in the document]
styfle: That said, it could be desirable to select an exact version. I remember a couple times where upgrading Node.js minor broke deployments. In particular when the “type” filed was added to package.json and tooling didn’t know about it yet.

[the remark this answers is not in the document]
wesleytodd: I am a bit confused about what we are saying is not true. I said "we need to support" not "everyone must do it this way". My point was just that the debians and docker containers and all that are still required no matter if we ship a version manager to help with local and ci dev.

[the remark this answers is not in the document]
GeoffreyBooth: > I am a bit confused about what we are saying is not true.

This part:

> in prod it is *always* necessary to support a secure and stable single version workflow.

sounds to me like you’re saying that the runtime, it not more, needs to be version-pinned in prod. Maybe you meant “pinned at the semver major level”? I read it as intending to mean pinned exactly.

coreybutler:
There is an inaccuracy in this under the technical details, or at least something that needs clarification. Most existing version managers do NOT manipulate the `PATH` for version switching. There are really only two strategies employed: shims and symlinks. The `PATH` is only used to identify the version manager binary, just like any other application. 

nvm-sh points to a shim, meaning a single executable is found in the `PATH`. The shim is responsible for executing the node binary. nvm-windows points the PATH to a "directory", which is a symlink. The symlink target is modified when the version changes, so the native node executable is always called directly.

We tried `PATH` manipulation in an early version of NVM for Windows. It was a _catastrophic_ failure. Windows can limit the size of the PATH, so changing it whenever a new version needs to be changed can truncate the `PATH`. If truncation occurred, at best user's other programs stopped working. At worst,  the entire OS crashed.

wesleytodd:
> Most existing version managers do NOT manipulate the PATH for version switching.

The most popular version manager [`nvm`](https://github.com/nvm-sh/nvm) uses `PATH` manipulation. Maybe we have unclear language here?

> nvm-sh points to a shim

I have it under very good authority that this is not the case, maybe you can explain a bit better what you mean here? do you mean the windows version and not what I linked above?

> We tried PATH manipulation in an early version of NVM for Windows. It was a catastrophic failure. 

I would love to hear more about this. Maybe you could join up on one of our meetings to share your experience with this?

coreybutler:
> The most popular version manager [nvm](https://github.com/nvm-sh/nvm) uses PATH manipulation.

I took a quick look at nvm's source and `PATH` is exported once in the `use` command, so it does appear to change the path whenever the version changes. I don't know if that's new or not (I'm sure Jordan could confirm). 

It's worth noting that nvm-sh is the most popular from a Github star perspective, but don't make the mistake of assuming it represents a majority of users or VMs. From the limited data we have, none of the individual VMs have a majority share of the overall user base.

Unless I'm mistaken, other VMs (like volta and nodist) use a shim as I described. Tools like n use the symlink approach.

There's not too much more to say about switching paths on Windows. Some versions of Windows limit the character count of the PATH. Truncate it and the system becomes volatile. The volatility can be unpredictable. If you append the node path to the end, it may be truncated and node won't be recognized. Prefix it and you risk cutting off system resource routes Windows needs to operate. Newer versions don't have this limitation, but there are a lot of people using older versions of Windows. There is also more than one `PATH` on Windows (system and user). They require different levels of permissions to modify. Throw in GPO's and many organizations explicitly block system/user env variable modifications. It's a can of worms.

wesleytodd:
> (I'm sure Jordan could confirm).

That is who messaged me about your comment :P

PR #597, opened 2024-04-10
Goals for Node.js and Package Manager Version Management Effort

GeoffreyBooth:
A draft of the goals discussed in https://github.com/nodejs/package-maintenance/issues/595.

cc @nodejs/package-maintenance @nodejs/tsc

### DIFF
--- a/docs/version-management/goals.md
+++ b/docs/version-management/goals.md
@@ -6,4 +6,4 @@ Help application developers to:
 
 1. Install Node.js and a package manager for a local development environment.
 
-1. Run different versions of Node.js and different versions of package managers for different projects.
+1. Run different versions of Node.js and different versions of package managers for different projects, deterministically.
--- a/docs/version-management/goals.md
+++ b/docs/version-management/goals.md
@@ -1,0 +1,9 @@
+# Goals for Node.js and Package Manager Version Management Effort
+
+Help application developers to:
+
+1. Define the correct Node.js runtime version and package manager version for projects.
+
+1. Install Node.js and a package manager for a local development environment.
+
+1. Run different versions of Node.js and different versions of package managers for different projects.
--- a/docs/version-management/goals.md
+++ b/docs/version-management/goals.md
@@ -6,4 +6,4 @@ Help application developers to:
 
 1. Install Node.js and a package manager for a local development environment.
 
-1. Run different versions of Node.js and different versions of package managers for different projects, deterministically.
+1. Run the correct version of Node.js and of a package manager for each project.


MoLow (approved):

docs/version-management/goals.md
```diff
@@ -0,0 +1,9 @@
+# Goals for Node.js and Package Manager Version Management Effort
+
+Help application developers to:
+
+1. Define the correct Node.js runtime version and package manager version for projects.
+
+1. Install Node.js and a package manager for a local development environment.
+
+1. Run different versions of Node.js and different versions of package managers for different projects.
```
mcollina, line 9:
```suggestion
1. Run different versions of Node.js and different versions of package managers for different projects, deterministically.
```

mcollina (commented):
I think it's missing the determinism.

docs/version-management/goals.md
wesleytodd: Yeah we had some language trying to address the "deterministic builds" part, we were a bit concerned because according to the folks we have talked with most folks are not pinning their version with `corepack` today. And I am not sure we want to "force determinism as a goal". We want to make it REALLY easy for folks to do, but our concern was that adding it to the goals might make it appear we would be committing to a technical decision with that language.

As long as folks realize that enabling but not forcing *is our goal*, then I am good with this change landing.

wesleytodd (approved):

docs/version-management/goals.md
aduh95: Realistically, we can't _force_ determinism. But we can make the deterministic way much easier and don't try to support the undeterministic way in the MVP.

docs/version-management/goals.md
```diff
@@ -0,0 +1,9 @@
+# Goals for Node.js and Package Manager Version Management Effort
+
+Help application developers to:
+
+1. Define the correct Node.js runtime version and package manager version for projects.
+
+1. Install Node.js and a package manager for a local development environment.
+
+1. Run different versions of Node.js and different versions of package managers for different projects, deterministically.
```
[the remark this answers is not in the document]
aduh95: I think having determinism is a goal, having a loose option is not a goal (but is also not off the table, it simply is not a priority)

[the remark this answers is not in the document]
GeoffreyBooth: Do we need to say anything at all about determinism? I think it’s implicit that we want to help users run the right software in the best way possible. I don’t think anyone would have read the original text and thought it meant that the runtime or package manager versions would be chosen randomly, for example.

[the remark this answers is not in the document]
GeoffreyBooth: I think this phrasing achieves the same?

```suggestion
1. Run the correct version of Node.js and of a package manager for each project.
```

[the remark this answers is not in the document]
wesleytodd: Yeah I personally think adding `determinism` implies more than we want for the high level goals and I think folks would read that and think we don't intend to support a loose mode. IMO we *must* support a loose mode from the start, but I dont want to get hung up on that here if we can help it. I am 👍 on either removing the new word *or* language like either I or @GeoffreyBooth  proposed above to clarify. I am 👎 on it as is now that I have thought about it for a bit.

wesleytodd (approved):

mhdawson (approved):
LGTM

docs/version-management/goals.md
```diff
@@ -0,0 +1,9 @@
+# Goals for Node.js and Package Manager Version Management Effort
+
+Help application developers to:
+
+1. Define the correct Node.js runtime version and package manager version for projects.
+
+1. Install Node.js and a package manager for a local development environment.
+
+1. Run the correct version of Node.js and of a package manager for each project.
```
styfle, line 9:
```suggestion
1. Define the correct Node.js runtime version and package manager version for each project.

1. Install the correct Node.js runtime version and package manager version for a local development environment.

1. Run the correct Node.js runtime version and package manager version for each project.
```

styfle: I wonder if we can make this much simpler:

- Define, run, and install the correct node version and package manager version

GeoffreyBooth: Well 'correct' only makes sense in the context of a project, right? But users might also want to install Node onto their machine generally, before creating any projects or without any particular project in mind.

joyeecheung:
I am not sure if this is the right place to ask, but I was struggling to find the right repo for this by googling and then I realized it is in the package maintenance working group and the README of this repo seems to have nothing to do with corepack or version management....or at least, we are talking about developing new features or new tools here, not talking about how to keep dead/old packages alive?

GeoffreyBooth:
> we are talking about developing new features or new tools here, not talking about how to keep dead/old packages alive?

Yeah, in the chartering discussion we wondered if perhaps it would be good to rename this team or start a new one. There’s a defunct @nodejs/version-management team too. I think for now the consensus was to deal with the goals first and then possibly rename/redefine the team based on what goals end up getting approved.

nullivex:
Looks good from my perspective.

docs/version-management/goals.md
```diff
@@ -0,0 +1,9 @@
+# Goals for Node.js and Package Manager Version Management Effort
+
+Help application developers to:
+
+1. Define the correct Node.js runtime version and package manager version for projects.
```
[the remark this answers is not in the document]
GeoffreyBooth: This document is intentionally omitting implementation details, because we want to define the goals without prescribing any particular solution.

bnb (approved):

docs/version-management/goals.md
styfle: Perhaps “specific”?

- Define, run, and install a specific node version and package manager version

ruyadorno (approved):

docs/version-management/goals.md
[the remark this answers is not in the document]
wesleytodd: @bnb in case you are wondering, the next step is to extract a bunch of things from this draft PR I opened with a running (and overly long) doc I had been compiling from these discussions. https://github.com/nodejs/package-maintenance/pull/594

joyeecheung:
Should I link to this PR in the trip report of the summit, or this repo? (somewhat weird to link to the repo while the REAMD still talks about package maintenance, but I guess we can also update the blog post)

joyeecheung:
Or I can link to https://github.com/nodejs/package-maintenance/issues/591 which should provide sufficient background

wesleytodd:
I wouldn't worry too much about the name. This group has always been a place to discuss ancillary tooling including package manager related topics. With this pivot we do need updates to the README, but as @GeoffreyBooth said we discussed chartering first and decided to establish the goals before chartering and changing the makeup of this group.

docs/version-management/goals.md
styfle: Sounds good. I just want to make sure we don't have implicit assumptions that are not listed here because that was the problem with previous solutions. For example, the goals here don't explain what "project" means.

marcelklehr:
:popcorn:

docs/version-management/goals.md
wesleytodd: Ah interesting point, yeah agreed we want to avoid assumptions. Should we list out "application or library" or do you have a particular assumption you think folks mean which we want to avoid folks making?

aduh95: Replacing "project" with "working directory" SGTM

GeoffreyBooth: I don’t think “working directory” makes sense. If I’m in `app/routes` and I have an `app/package.json`, it’s the `package.json` root that defines the project scope; I shouldn’t need another `app/routes/package.json` to repeat the version of Node to use for commands in `app/routes`. I know we might not define configuration based on `package.json`, but I also think it’s unlikely that we would want to define it separately in every folder within a project.

aduh95: Saying "Define the correct Node.js runtime version and package manager version for the current working directory" says nothing about having the config file in the same directory, it could be in a different one (the parent directory in your example)

GeoffreyBooth: It implies that the unit we’re concerned with is the directory, when really what we’re focusing on is projects. Our goal here is to help users run the right software for their projects.

wesleytodd (approved):

docs/version-management/goals.md
Eomm, line 5: I think the only `correct` version is LTS, but it is there are tons of projects that work with EOL Node.js, so I would say the `needed` or `desired`

```suggestion
1. Define the needed Node.js runtime version and package manager version for projects.
```

GeoffreyBooth: I don’t know of anywhere that we say that non-LTS versions are not “correct.” They just don’t have long-term support.

PR #598, opened 2024-04-26
doc: propose revising the downloads page

GeoffreyBooth:
Following up #597, this is a proposal for a first step toward achieving those goals.

cc @nodejs/package-maintenance @nodejs/nodejs-website @nodejs/tsc

### DIFF
--- a/docs/version-management/proposal-revise-downloads-page.md
+++ b/docs/version-management/proposal-revise-downloads-page.md
@@ -1,0 +1,13 @@
+# Proposal: Revise Downloads Page
+
+As part of achieving the second [goal](./goals.md):
+
+> Install Node.js and a package manager for a local development environment.
+
+Our view is that the optimal way to install Node.js and bootstrap components like a package manager is through one or more external tools.
+
+As an initial step towards this, we should revise the Node.js download page, https://nodejs.org/en/download, so that the primary download instructions suggest installation via one or more tools that provide for managing the versions of the Node.js runtime and of a package manager.
+
+The current primary download method, a link to an installer, would still be an option available in one of the download page’s tabs; but it would no longer be the default method. This is because the installer link is only appropriate for certain target users, such as those who use Node.js for running scripts, and not for our primary target user group of application developers. Our default recommendation should be targeting our default set of users, and therefore it should recommend installing via one or more tools that make it easy to manage the runtime and package manager per project.
+
+We should do this revision now, without predicating it on any changes to the Node.js distribution or on creating new tools. Once the page is rewritten, hopefully it should be apparent where any distribution changes or new tools might provide for an improved developer experience, and we can pursue ideas in those directions.


targos:
I think I generally agree with that. There are issues with external tools, though. For example Chocolatey takes time to update.
Versions 22.0.0 and 21.7.3 were only added yesterday, and version 21.7.2 was apparently skipped:
https://community.chocolatey.org/packages/nodejs#versionhistory
CleanShot 2024-04-26 at 11 16 08

targos:
Main problem being: on the website we show the command `choco install nodejs --version="22.0.0"`. This command is guaranteed to fail until the chocolatey distribution is updated.

richardlau:
> For example Chocolatey takes time to update. Versions 22.0.0 and 21.7.3 were only added yesterday, and version 21.7.2 was apparently skipped

FTR Chocolatey used to scrape https://nodejs.org/en/download and https://nodejs.org/en/download/current, which [broke when the website redesign went live](https://github.com/orgs/nodejs/discussions/52407). https://github.com/chocolatey-community/chocolatey-packages/pull/2453 was merged yesterday.

ovflowd:
I assume then that this is a self-fulfilling prophecy? It should not work out of the box?

wesleytodd:
If they need a tool to scrape the updated data generated by the build team we have one maintained by this groups (the PMWG) here: https://github.com/pkgjs/nv

I agree this is an issue and any tools we recommend on the page should have a fast and reliable support policy. Do we need to add language clarifying the requirements to be included on the download page maybe?

richardlau:
> If they need a tool to scrape the updated data generated by the build team we have one maintained by this groups (the PMWG) here: https://github.com/pkgjs/nv
> 

I did point the person asking about this to that 🙂: https://github.com/orgs/nodejs/discussions/52407#discussioncomment-9038845

michha:
Just some short input from me (who repaired the choco script for nodejs):
- I had just the urge to get this fixed because I needed the package for myself (it was quite a challenge to get to the reason why the choco package versions were absent)
- from the .NET community I am familiar with their [releases-index.json](https://[dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json](https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json)
- dealing with the pointed out tool (pkgjs/nv) would have lasted even longer (and I even cant say if using another tool instead of plain PowerShell would be appreciated by the choco package repo owner), so I did the parsing by myself, which was quite easy
- the current solution should be pretty stable as long as the location of json files isnt changed 😅
- I dont know how often the script is executed but it seems several times per day, so new versions should be picked up within 24 hours of publishing inside the json files

wesleytodd:
> the current solution should be pretty stable as long as the location of json files isnt changed 

Just to be particular about this, the reason we thought it was a good idea for the node project to own a tool for this job was so the build or release group *could* change the location and have a clear and known place to help the community keep up. I don't know anything about the choco package or the maintainer, so I trust you made a good decision here, but IMO the "it shouldn't move" is very near to "it shouldnt change" which is equally problematic from a contract perspective. It is an un-versioned resource, which leads to issues like this. Luckily it is a low change thing so likely not a big deal, but if you can suggest to them that "in order to remain on the download page instructions for the project it is *recommended* to use their tool for fetching and parsing the data that would be idea IMO.

GeoffreyBooth:
All this discussion of Chocolatey is off-topic, in my opinion. Nothing in this PR mentions Chocolatey or any particular package or version manager. This proposal is just about the _idea_ of revising the downloads page toward a new general goal, not the _particulars_ of what the new draft should be. I think we can get into those details in the PR that actually does the rewrite, assuming that this proposal is approved.

wesleytodd (approved):

wesleytodd:
Yeah, I do think that we should address at some point (maybe not this point) what the qualities we require things to have to be included in those instructions for the future. But for now I agree it is likely not required to land this PR and then do the website changes.

mhdawson (approved):
LGTM

mhdawson:
I agree that the discussion of specific tools is off topic for this PR.

I think if we agree on the `approach` then the specific issues and problems with what we might recommend can be handled in the PRs and discussion around what to recommend.

The approach does not pre-suppose any solution as the target end state. It `might` be to use existing tools, it `might` be to create a new tool that is cross platoform.

There will likely be challenges with existing tools which is what are likley to be documented in the recommended first step of updating the web site, but even if they are imperfect they might be a reasonable step towards the longer term goal.

ronag (approved):

styfle (approved):

larson-carter (approved):
LGTM?

PR #600, opened 2024-05-09
doc: add minutes for meeting 9 May 2024

mhdawson:


mhdawson:
@nodejs/package-maintenance would be great to get a review so that this could land.

thescientist13 (approved):

PR #617, opened 2024-10-22
docs: add @bjohansebas

bjohansebas:
I am currently working on getting the statusboard to work with Express again, and I would like to contribute to the statusboard since it is not currently functioning with recent versions of Node.js.

### DIFF
--- a/MEMBERS.md
+++ b/MEMBERS.md
@@ -44,3 +44,4 @@
 * Luke Karrys ([@lukekarrys](https://github.com/lukekarrys))
 * Claudio Wunder ([@ovflowd](https://github.com/ovflowd))
 * Leo Balter ([@leobalter](https://github.com/leobalter))
+* Sebastian Beltran ([@bjohansebas](https://github.com/bjohansebas))


wesleytodd:
We need to probably figure out how we want to address these kinds of things. @bjohansebas has been a great new contributor over in Express, but this group is governed by Node.js and the PMWG. In the past I would have said we keep participation in the PMWG as the requirement for contributing to the pkgjs org, but I am not so sure anymore since a lot of packages in there are not well maintained.

For now @bjohansebas I think the best course is to just open PRs like you have been and I will try and address them more quickly. But I think this is something we should get on the agenda for the next PMWG meeting to discuss.

bjohansebas:
Beyond pkgs and the statusboard, I would like to be closer to the discussions happening with this group.

wesleytodd:
Yep! Makes sense. I was more thinking that we should loosen the guidelines so that we have some mechinism in place to add committers who didn't really participate in the discussions. Great that you want to participate more in the group though. Ill merge this, sorry it has sat so long.

bjohansebas:
Since I want to participate more in the groups, maybe I could be added to the working group (`@nodejs/package-maintenance`) to receive the notifications.

